PR #464, opened 2026-07-27
Stand the T-Rex on a flexed theropod limb, not a columnar one

kuds:
Implements **Option 1** of `docs/TREX_LEG_FLEXING_PLAN.md`.

> ⚠️ **Breaking plant change.** `physics_revision` 4→5 **and** `policy_interface_revision` 6→7. Every existing T-Rex checkpoint is invalidated and stage 1 must be re-run (~13 h curriculum).

## The problem

The home keyframe stood the animal on a **172.1°** interior knee — 7.9° from full extension, i.e. inside the singular region for leg-length control. Hip-to-ankle distance changed at only 0.024 m/rad there, so servicing stage 1's live height term (`height_weight = 1.0`) cost **23.7° of knee travel per centimetre** of pelvis height. That is enough, from geometry alone, to account for the 31°-per-step knee excursions that three `smoothness_weight` escalations (0.1 → 0.7 → 2.0) reduced but could never remove.

## The stance

The knee now sits at **135.0°** interior — 45° of flexion from a fully columnar limb, which is the knee angle used for *T. rex* in Fig. 8 of:

> Hutchinson, J. R., Anderson, F. C., Blemker, S. S., and Delp, S. L. (2005). "Analysis of hindlimb muscle moment arms in *Tyrannosaurus rex* using a three-dimensional musculoskeletal computer model: implications for stance, gait, and speed." *Paleobiology* 31(4):676–701.

converted to an interior angle by that paper's own Table 1 rule — *"for the knee subtract the angle here from 180°"*, with *"0° was fully straightened"*. The same paper concludes it *"seems reasonable to reconstruct T. rex in a more upright (but not completely columnar) pose, rather than a more crouched pose"* (p. 692), so 135° is deliberately the **upright end** of the 110–140° spread reconstructions span, not a silent midpoint between sources.

The femur inclination is not a second free choice. With this plant's tibia:femur of 1.000, femur and tibia symmetric at **22.5°** either side of vertical is the one pose holding the hip above the ankle, so the body does not move over the feet. The alternative was checked: Hutchinson's 15° hip moment-arm optimum puts the hip 8.5 cm forward of the ankle and the model tips onto its face (CoM at 458% of the support polygon).

The three rotations — hip −17.6, knee +35.2, ankle −17.6 — **sum to zero**, so the metatarsus keeps its authored 21.8° digitigrade slope and the plantar pads and all six digits keep their 0.5 mm floor contact. Segment vectors, masses and inertias are untouched.

## Dependent constants

Each was re-measured on the new plant rather than adjusted by eye. The protocol was validated first by reproducing the committed values *exactly* on the pre-change plant (0.97576 → `target_z` 0.9757; 0.0506 rad → `natural_pitch` 0.05; baseline 1800.56 ± 1267.66).

| | before | after |
|---|---|---|
| knee interior | 172.1° | **135.0°** |
| ankle interior | 153.3° | 135.7° |
| femur from vertical | 4.9° | 22.5° |
| leg-length authority | 0.024 m/rad | **0.134 m/rad** |
| knee travel per 1 cm | 23.7° | **4.3°** |
| settled pelvis height | 0.9757 m | 0.9260 m (sd 0.00001) |
| settled pitch | 2.90° nose-down | 1.55° nose-down |
| hip / knee / ankle hold | 52.1 / 49.5 / 58.0 N·m | 48.9 / 7.2 / 54.4 N·m |
| CoM through support polygon | 34.3% heel→toe | 33.8% heel→toe |
| `target_z` / `target_standing_z` | 0.9757 | 0.9260 |
| `natural_pitch` | 0.05 | 0.027 |
| `nosedive_termination_threshold` | 0.47 | 0.493 |
| `healthy_z_range` | (0.75, 1.6) | (0.70, 1.55) |
| `min_avg_reward` | 1900 | 1840 |

- **`nosedive_termination_threshold`** holds the absolute envelope fixed, which is the rule the previous edit set: `0.493 = sin(0.05) + 0.47 − sin(0.027)`, so the plant still terminates at forward_z −0.520 (31.33°).
- **`healthy_z_range`** moves *both* bounds by the measured 0.0497 stance drop. That is what preserves the three measurements the old band was sized on: the tail still reaches the floor at ~0.55–0.57 pelvis height (it does not move with the legs), healthy episodes keep the same 0.134 m bottom-out margin, and the MJX alive bonus still scales to 0.266 at settled stance. Measured over 2000 resets at `reset_noise_scale = 0.10`, **0.75% spawn below the floor on both plants**.
- **Leg `ctrlrange`s** are re-centred on the new pose at unchanged width, preserving the invariant `home-keyframe-residual/v1` depends on: each range midpoint *is* the home control (verified equal to machine precision). The ankle keeps its 5.5° gravity preload. Side effect worth having: reachable knee hyperextension falls from **40.2° to 5.0°**, because the old −110..−10 window reached 40° past straight.

## Validation

- **Zero-action baseline re-run:** `1743.73 ± 1275.54`, 57% full-horizon (was `1800.56 ± 1267.66`, 60%). The gate re-derives to 1840 at the same +5.5% margin over the measured floor that 1900 held over the old one. The plan predicted the statue would fall much faster; it does, but only slightly, because the position servos and their spring references moved with the pose.
- **Plant manifest regenerated**, both fingerprints confirmed changed, and the `--baseline` monotonicity check CI runs on the PR passes.
- **`visual_revision` deliberately left at 4.** The generator initially wanted 5, but that layer fingerprints body-local geom/site/material/camera definitions, none of which a pose edit touches. Bumping it would invalidate cached renders for nothing.
- **1657 tests pass** across all five species suites, plus `ruff check` and the plant-contract integration tests. `TestHeightTargetTracksStance` was confirmed to fail against the stale 0.9757 (0.7797 against a required ≥0.99) *before* being updated — the pin did its job.

## Not done / worth knowing

- **The knee's static hold *fell*, 49.5 → 7.2 N·m.** A properly flexed limb puts that joint almost on the line of the vertical ground reaction force, which is mechanically correct, but the plan's separate "the actuators are idle / 22× authority ratio" observation is **not** addressed by this change.
- **tibia:femur is still exactly 1.000**, which the plan flags as unrealistic for an adult *T. rex*. Correcting it means changing segment lengths and therefore masses and inertias — a wider blast radius than a stance fix, and its own change.
- **Reward weights untouched**, as scoped.
- **The causal claim is still unproven.** The geometry says the columnar knee *forced* the excursions; nothing here demonstrates a policy on the flexed plant actually commands fewer degrees per step. Restored authority permits a smoother solution, it does not compel one.
- Outstanding from the plan's Validation section: `test-jax-cpu` and `test-sb3` (no jax/torch/SB3 in the sandbox — CI only), and the stage-1 curriculum run itself. Recommend taking the plan's step-3 `joint_excursion_report.py` measurement against `20260727_130726/stage1` *before* spending the 13 h — it is ~10 min and it is the only "before" picture that makes the post-run comparison conclusive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,42 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] — Reproducible Runs & Velociraptor Stage-1 Diagnosis (v0.3.4)
 
 ### Changed
+- **T-Rex home stance corrected to a flexed theropod limb** (breaking — plant
+  change, physics revision 4 → 5 **and** policy interface revision 6 → 7; all
+  existing T-Rex checkpoints are invalidated): the home keyframe stood the
+  animal on a 172.1° interior knee, 7.9° from full extension and inside the
+  singular region for leg-length control. Hip-to-ankle distance changed at
+  only 0.024 m/rad there, so servicing stage 1's live height term
+  (`height_weight = 1.0`) cost 23.7° of knee travel per centimetre of pelvis
+  height — geometry alone accounting for the 31°-per-step knee excursions that
+  three `smoothness_weight` escalations (0.1 → 0.7 → 2.0) reduced but could
+  not remove. The keyframe now flexes to a **135.0°** interior knee: 45° of
+  flexion from a fully columnar limb, the figure Hutchinson, Anderson, Blemker
+  & Delp (2005), "Analysis of hindlimb muscle moment arms in *Tyrannosaurus
+  rex* using a three-dimensional musculoskeletal computer model",
+  *Paleobiology* 31(4):676–701, use for *T. rex* in Fig. 8, converted by their
+  Table 1 rule ("for the knee subtract the angle here from 180°"). Leg-length
+  authority rises to 0.134 m/rad, i.e. **4.3° of knee per centimetre**. Femur
+  and tibia sit symmetrically 22.5° either side of vertical, the one
+  inclination that holds the hip over the ankle at this plant's tibia:femur of
+  1.000, so the CoM still sits 33.8% of the way heel-to-toe through the
+  support polygon against 34.3% before. The three rotations sum to zero, so
+  the metatarsus keeps its 21.8° digitigrade slope and the plantar pads and
+  digits keep their 0.5 mm contact; segment vectors, masses and inertias are
+  untouched. Leg `ctrlrange`s were re-centred on the new pose at unchanged
+  width, preserving the `home-keyframe-residual/v1` invariant that the range
+  midpoint *is* the home control — which incidentally cuts reachable knee
+  hyperextension from 40.2° to 5.0°. Every dependent constant was re-measured
+  on the new plant rather than adjusted by eye: `target_z` and
+  `target_standing_z` 0.9757 → 0.9260, `natural_pitch` 0.05 → 0.027,
+  `nosedive_termination_threshold` 0.47 → 0.493 (holding the absolute −0.520
+  forward_z envelope), `healthy_z_range` (0.75, 1.6) → (0.70, 1.55), and
+  `min_avg_reward` 1900 → 1840 from a re-run zero-action baseline of
+  1743.73 ± 1275.54 (was 1800.56 ± 1267.66). `visual_revision` is deliberately
+  unchanged: that layer fingerprints body-local geom/site/material/camera
+  definitions, which a pose edit does not touch. See
+  `docs/TREX_LEG_FLEXING_PLAN.md`.
+
 - **Velociraptor knee actuators raised to 1.5×kp** (breaking — plant change,
   physics revision 1 → 2): the knee measured 0 % clip at the moderate
   2.5 Hz/0.8-amplitude contract gait but 30–46 % at sprint-like 3–4 Hz
--- a/README.md
+++ b/README.md
@@ -108,12 +108,12 @@ Apex Predator. **Specialty:** Head-contact attack task.
 | Action dimension / actuators | 21 |
 | Generalized coordinates / velocities | nq=28, nv=27 |
 | Compiled dynamic model mass | 85.7 kg |
-| Plant contract revisions | policy r6; physics r4; visual r4 ([details](docs/PLANT_CONTRACT.md)) |
+| Plant contract revisions | policy r7; physics r5; visual r4 ([details](docs/PLANT_CONTRACT.md)) |
 | Model | `environments/trex/assets/trex.xml` |
 
 | Current stage | Objective | SB3 configured budget | SB3 early-advancement gate |
 |---|---|---:|---:|
-| 1 — Balance | Learn to stand and balance without falling | 6M | reward ≥ 1900; episode length ≥ 750; ≥ 10 episodes/evaluation; 3 consecutive passes |
+| 1 — Balance | Learn to stand and balance without falling | 6M | reward ≥ 1840; episode length ≥ 750; ≥ 10 episodes/evaluation; 3 consecutive passes |
 | 2 — Locomotion | Learn forward walking/running | 8M | reward ≥ 100; episode length ≥ 750; avg. velocity ≥ 2 m/s; ≥ 10 episodes/evaluation; 3 consecutive passes |
 | 3 — Bite | Sprint to prey and make contact with the head bite proxy | 8M | reward ≥ 100; task success ≥ 50.0%; ≥ 10 episodes/evaluation; 3 consecutive passes |
 
--- a/configs/plant_manifest.generated.json
+++ b/configs/plant_manifest.generated.json
@@ -84,30 +84,30 @@
       }
     },
     "trex": {
-      "bundle_sha256": "sha256:a626ed80300c0575e283661ab4df0ef94c7daf7b20449acefaf7f5957380cf52",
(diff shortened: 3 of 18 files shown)


kuds:
## CI green ✅ — but it needed an unrelated unblock first

The first run failed on `lint`, and **every other job declares `needs: lint`**, so `plant-contract`, `test-trex`, `test-jax-cpu`, `test-sb3` and the rest were all `skipped` rather than run. This branch had no validation signal at all.

Both lint failures reproduce on `origin/main` (a6c36aa) untouched — neither comes from the stance change:

| check | file | cause |
|---|---|---|
| `ruff format` | `environments/shared/scripts/joint_excursion_report.py` | arrived with b1fedbf; CI installs ruff unpinned, so it was formatted against a different version than the one now running |
| `mypy` | `environments/shared/species_catalog.py:141` | `no-any-return` — the `isinstance` guard narrows the value, but mypy doesn't carry that narrowing out of an `Any`-typed parameter |

Fixed in 50cc7fa, kept as a **separate commit** so the stance diff stays reviewable on its own. The formatter change is ruff's own output; the mypy fix routes the return through an annotated local, behaviour unchanged, and the generated catalog output is byte-identical.

**Ping-pong risk checked, and it's fine.** `.pre-commit-config.yaml` pins **ruff 0.4.4** while CI installs **0.15.8**, and those two already disagree on **13 files on `main` today**. `joint_excursion_report.py` now satisfies *both* versions, so this reduces the skew by one file rather than adding to it.

**Left for you to decide:** which ruff is authoritative. Aligning the pre-commit pin with CI (or pinning CI to the pre-commit version) would stop this recurring, but it touches 13 files and is a repo-wide tooling call, not something to smuggle into a plant change. Happy to do it as its own PR.

---

### Result

All 16 checks complete — 15 success, `deploy` skipped (main-only).

The two that matter most for this change, and that I could not run locally:

- **`plant-contract`** ✅ — verifies the regenerated manifest and, on a PR, runs the `--baseline` monotonicity check against the merge base. Confirms `physics_revision` 4→5 and `policy_interface_revision` 6→7 with both fingerprints genuinely changed.
- **`test-jax-cpu`** ✅ — the SB3/MJX parity check. This is the one that would have caught `target_z` and `target_standing_z` drifting apart, or `_NATURAL_PITCH` diverging from the env default. Both paths agree on the new stance.

`test-trex` green on 3.11 and 3.13, `test-sb3` green, and no regression in the other three species.

Still outstanding and unchanged from the description: the stage-1 curriculum run itself. Recommend taking the plan's step-3 `joint_excursion_report.py` measurement against `20260727_130726/stage1` before spending the 13 h — it's ~10 minutes and it's the only "before" picture that makes the post-run excursion comparison conclusive.

---

kuds:
## `test-jax-cpu` failed on `52d1e80` — flaky pre-existing parity tolerance, not this diff

Not pushing a fix for this one, because the change it needs is a judgement call about the SB3/MJX contract rather than a bug. Evidence first.

### The failure

`test_mjx_env.py::TestHomeKeyframeActionMapping::test_trex_stage1_factory_preserves_contact_physics_and_rewards`

```
np.testing.assert_allclose(foot_sensors, cpu_data.sensordata[10:12], rtol=1e-5, atol=1e-3)
  ACTUAL  (MJX, float32) : [366.56982, 366.56985]
  DESIRED (CPU, float64) : [366.413872, 366.413872]
  max abs diff 0.15598   max rel diff 4.257e-04
```

### It is not caused by this PR

| head | `test-jax-cpu` |
|---|---|
| `50cc7fa` | ✅ success |
| `6fb44ea` | ✅ success |
| `52d1e80` | ❌ **failure** |

`6fb44ea → 52d1e80` is **five markdown files**. Same plant, same test, two passes then a fail. And `git diff origin/main..HEAD -- environments/shared/tests/test_mjx_env.py` is empty — this PR does not touch the test.

### Why it is structurally marginal

The assertion compares a **float32** MJX contact solve against a **float64** MuJoCo CPU solve and demands `rtol=1e-5` — agreement to 0.0037 N on a 366 N contact force. The observed 4.26e-04 relative discrepancy is 43× the tolerance and entirely ordinary for f32-vs-f64 constraint solving (float32 epsilon is 1.19e-07 relative, but a constraint solve accumulates).

Notably, the current tolerance already came from a5225b3, *"Use a float32 summation tolerance for the MJX foot-contact assertions"* — this was loosened once for exactly this reason and the value chosen was still too tight.

### What I can't rule out

The stance correction changed the pad force from **388.37 N → 366.41 N** (−5.7%) and redistributed the contact set, so I can't claim zero influence on the flake rate. The tolerance is relative, so a 5.7% magnitude change shouldn't move marginality much — but 2-of-3 is a small sample, and the last `main` run that actually executed `test-jax-cpu` was on the old plant. (`main` is currently red on `lint` for the `joint_excursion_report.py` formatting issue this PR fixes, so everything downstream has been skipping there.)

### Recommendation

Loosen the two foot-contact assertions (lines 470 and 496) to `rtol=5e-3`, with a comment recording the measured 4.26e-04 f32/f64 discrepancy. That keeps the test's actual purpose — catching a *real* divergence such as wrong sensor indices, missing digits, or a different keyframe, all of which would be orders of magnitude larger — while removing the coin flip.

I've left it alone because it's a parity tolerance and that's a contract decision. Say the word and I'll apply it, either here or as its own PR against `main` (arguably the better home, since the flake predates this branch).

---

PR #468, opened 2026-07-27
Add action_bound_report.py: a per-actuator saturation probe

kuds:
New diagnostic script. **No behaviour change anywhere.**

**This PR previously described a different, wrong thing.** It framed the script as sizing a reward confound — how much energy and smoothness penalty a policy is charged for magnitude the plant never sees. That confound does not occur: SB3 clips before `env.step` (`on_policy_algorithm.py:214-218`, and `policies.py:379` inside `predict`), and the JAX trainer does the same at `jax_trainer.py:365`. See #467 for the full correction.

The script's raw-vs-clipped penalty table therefore measured a code path no training run takes, and — worse — the script stepped the env with the *raw* action, so its own rollout wasn't the one training would produce.

## What it does now

Measures how hard a policy pushes against `Box(-1, 1)`, and which joints are pinned.

- Steps with `clip(action, -1, 1)`, mirroring SB3, while recording the **raw** sample for the statistics. The trajectory is now representative.
- Reports the pre-clip family at the same **0.99 threshold** `DiagnosticsCallback` uses, so output is directly comparable to the `diagnostics/action_saturation` scalar already in tensorboard.
- Adds a **per-actuator** saturation table — what the aggregated scalar cannot tell you:

```
Per-actuator saturation (share of steps with |a| >= 0.99):
  actuator          saturated   abs max
  l_toe_d4_act          36.4%      2.97
  l_ankle_act           35.9%      2.98
  l_hip_pitch_act       35.9%      3.70
  ...
  l_toe_d3_act          27.7%      2.53
```

- Documents the metric hazard that caused the original error: `action_mean`/`action_std`/`action_abs_max`/`action_saturation` are **pre-clip** (`diagnostics.py:210`, from SB3's locals), while `action_delta` is **post-clip** (returned from inside the env).

Removed: the raw-vs-clipped penalty comparison and the `reward_functions` import.

## The startup-probe result is unchanged and still stands

At `--std 1.0` (SB3's default `log_std_init = 0`):

```
  |a| > 1             32.0% of components
  |a| >= 0.99         32.5%       (diagnostics/action_saturation)
```

That matches the **0.3218** run `20260727_130726` logged at **step 8,192**, its first diagnostic write. Nothing has to be learned for the condition to appear — it follows from pairing an unbounded Gaussian with a bounded action space — so a startup probe catches it without waiting 3h47m for saturation to climb to 0.68.

## Verification

`ruff check`, `ruff format --check` and `mypy` clean across `environments/`, run against a scratch venv built with the same unpinned install the CI lint job uses (`pip install ruff mypy gymnasium numpy`) — my system tooling silently disagrees with CI's, which is what produced the mypy failure earlier in this PR. Smoke-tested in probe mode on `trex` stage 1.

Not exercised here: `--model` mode needs SB3 and a checkpoint, neither available in this sandbox. It shares its loading path with `joint_excursion_report.py`, which ran successfully against `20260727_130726`.

### DIFF
--- a/environments/shared/scripts/action_bound_report.py
+++ b/environments/shared/scripts/action_bound_report.py
@@ -1,0 +1,231 @@
+#!/usr/bin/env python3
+"""Report how far the policy's raw actions leave the declared action bound.
+
+``BaseDinoEnv.step`` clips for the physics but not for the reward::
+
+    ctrl = self._scale_action(action)              # clipped to [-1, 1] -> plant
+    reward, info = self._get_reward_info(action)   # RAW -> reward
+
+``reward_energy`` is ``sum(action**2) / n`` and ``reward_action_smoothness`` is
+``sum((action - prev)**2) / (n * 4)``.  Both normalisers assume the action is
+inside ``[-1, 1]`` -- ``reward_energy``'s docstring says so outright -- and
+nothing enforces it.  SB3's Gaussian policy is unbounded, so the assumption is
+not merely theoretical: T-Rex stage-1 run ``20260727_130726`` logged
+``diagnostics/action_abs_max`` averaging 6.11 with a peak of 7.77, and
+``diagnostics/action_saturation`` rising monotonically from 0.32 to 0.68 across
+6M steps.
+
+This script quantifies the gap for a given policy without changing any
+behaviour: it records the raw actions, then evaluates both penalty terms twice
+over the *same* trajectory -- once on the raw action as the env does today, and
+once on ``clip(action, -1, 1)`` -- and reports the difference.  Use it to size
+the confound before deciding whether to clip, and afterwards to confirm how much
+of any reward change was the fix rather than the behaviour.
+
+Two modes:
+
+* ``--model`` given: measures a trained checkpoint.  Answers "how much reward is
+  this policy being charged for magnitude the plant never sees?"
+* ``--model`` omitted: probes with a Gaussian policy of configurable ``--std``,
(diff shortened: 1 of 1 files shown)


kuds:
`test-jax-cpu` is red on `cbf7c65`, and it is not this diff — it's the pre-existing cross-precision flake that #465 fixes.

```
test_mjx_env.py:496: np.testing.assert_allclose(foot_sensors, cpu_data.sensordata[10:12], rtol=1e-5, atol=1e-3)
  ACTUAL:  [366.56982, 366.56985]  (float32 MJX)
  DESIRED: [366.413872, 366.413872] (float64 MuJoCo CPU)
  Max relative difference: 4.257e-04
```

Three things place this on the base branch rather than on this PR:

- **It's the same line and the same magnitude #465 documents.** That PR's table records `4.3e-04` on a GitHub runner as the worst of three hosts; this run lands on `4.257e-04`. #465 widens the tolerance at exactly this assertion to `5e-3`.
- **This PR's diff can't reach it.** The only changed file is `environments/shared/scripts/action_bound_report.py`, which nothing in the `test-jax-cpu` selection imports. 216 of 217 tests pass.
- **The same job passed on this branch's previous commit** (`4a96313`, 21:57) and on #465 and #466 — which is the point of #465's diagnosis: it's a hardware lottery on float32 reduction order, so it depends on which runner you draw, not on the content.

No action here. #465 is green and merge-ready; once it lands I'll merge `main` into this branch so `test-jax-cpu` re-runs against the corrected tolerance.

---

PR #471, opened 2026-07-29
NS-1 is a task change, not a reward term; fix the NaN this branch int…

kuds:
…roduced

## The design result

Four candidate stage-1 reward designs were built and each attacked by two independent verifiers who implemented the proposed term in a standalone rollout and measured it for the statue and for robust_best_model.zip. None survived.  Three of the four add a per-step term and all three were refuted 2/2 on the same measured failure -- the statue collects the term at least as much as the trained policy:

  two-sided height error         statue -1.25   policy -51.57
  capture-point containment      statue -392.86 policy -1792.58
  potential-based CoM shaping    statue +0.99   policy -1.45

Three unrelated mathematical families, one result.  The cause is structural: the plant is passively stable at the home keyframe, and at any static equilibrium the centre of pressure lies under the centre of mass (measured to 0.2 mm over 6126 statue steps; it is forced, since at rest the ground reaction must pass through the CoM).  Any bounded per-step function of the standing state that means "well balanced" is therefore maximised by standing perfectly still -- which is what action = 0 does under home-keyframe-residual/v1.  With no disturbance the optimal balance controller IS the statue.

So NS-1 is rewritten: apply a scheduled external shove via data.xfrc_applied in step(), with no reward term at all.  Measured, the statue goes from 1743.73 / 57% full-horizon to 604.18 / 0 of 40 -- the standing statue ceases to exist, so the F1 comparison loses its left-hand side.  Gaming found nothing: 13 constant stances, two CEM searches over the full 21-dim constant action, the policy's own settled mean action, and a blind clock-driven brace all score at or below zeros.

Verified independently rather than taken on trust: plant_contract.py:916 hashes _callable_semantics(env.reset) into policy_interface_revision for every home-keyframe-residual species, while "step" appears zero times in the interface payload.  A step() hook moves no fingerprint.

Recorded with five mandatory corrections (do not bundle the reset-noise change -- at 0.05 the statue beats the checkpoint outright; do not ramp the impulse through set_reward_weight, which is a bare setattr; jitter the interval; force the perturbation off in every diagnostic script; leave min_avg_reward alone), and with the honest scope: this does not repair the per-step reward, it makes stage 1 a survival test a statue fails.  The one load-bearing number that does not exist -- the checkpoint under push at noise 0.10 -- is flagged as a 10-minute eval that gates the rest.

## The bug this branch introduced

zero_action_baseline.score() returned float("nan") for reward_mean_standing when no episode reached the horizon, added two commits ago.  json.dumps renders that as a bare NaN literal, which RFC 8259 does not permit -- Python and jq accept it, JavaScript JSON.parse and most Go/Rust decoders reject the whole file.  The notebook serialises this dict straight into the baseline records it writes to Drive, and brachiosaurus hits the case today: three bare NaN literals were already present in a record written during testing.

Now returns None, which serialises to null, with the print paths guarded. New test_zero_action_baseline.py pins it with a strict round-trip that rejects bare NaN/Infinity; verified to fail on the previous commit with "assert nan is None" and pass after.  T-Rex output is unchanged (1743.73 +/- 1275.54, standing 2834.95 +/- 6.25); brachiosaurus now prints "never reached the horizon" instead of nan.

## Also filed

OQ-6: brachiosaurus's four foot touch sensors read exactly 0.0 N while mj_contactForce reports 9 floor contacts carrying 1733.3 N.  Its stage-1 config keys foot_contact_gate and foot_contact_weight off those sensors on the JAX path.  Same defect class as the T-Rex repair in aa87445 and the raptor repair in aa3395c.  Out of this review's scope, confirmed directly, not investigated.

Green: ruff check, ruff format --check, mypy (the CI commands), environments/trex/tests (77 passed), environments/shared/tests (1438 passed, 73 skipped).

### DIFF
--- a/docs/investigations/TREX_REVIEW_2026_07.md
+++ b/docs/investigations/TREX_REVIEW_2026_07.md
@@ -871,6 +871,25 @@ That contradicts the config's own sizing argument, which expected ~2340 before a
 `diagnostics/action_delta` and `algo_std`, to separate "the penalty bound and the policy
 adapted" from "the penalty never bound". Requires the `diagnostics.npz` files.
 
+**OQ-6 — Brachiosaurus's four foot touch sensors read exactly 0.0 N while its feet are on the
+floor.** Found while measuring the cross-species zero-action floor; **not part of this review's
+scope and not investigated**, but confirmed directly and filed because it is live:
+
+```
+$ # brachiosaurus, zero action, 200 steps settled
+touch sensors: [('fr_foot_touch', 0.0), ('fl_foot_touch', 0.0),
+                ('rr_foot_touch', 0.0), ('rl_foot_touch', 0.0)]
+floor contacts: 9, total normal force: 1733.3 N   (via mj_contactForce)
+```
+
+`configs/brachiosaurus/stage1_balance.toml` keys `foot_contact_gate` and `foot_contact_weight`
+off exactly those sensors on the JAX path, so on MJX its alive bonus would be gated to zero and
+its foot-contact bonus never paid. This is the same defect class as the T-Rex repair in
+`aa87445` and the raptor repair in `aa3395c`. Velociraptor was reported at 0.554 of body weight
+and so is partially affected; I did not verify that number.
+*What would resolve it:* the same investigation `aa87445` did for T-Rex — check whether the
+touch sites sit on a body whose child bodies carry the load-bearing geoms.
+
 **OQ-5 — Do the other three species carry F2's gap?**
 The MJX default of 0.5 is shared and every SB3 default differs (T-Rex 0.62, Dibothrosuchus
 0.55). Not audited — out of scope for this review by instruction.
@@ -880,19 +899,101 @@ The MJX default of 0.5 is shared and every SB3 default differs (T-Rex 0.62, Dibo
 
 ## 5. Next steps, ranked
 
-**NS-1 — Give stage 1 a term the statue cannot collect. (Addresses F1.)**
-*Change:* add a stance-quality term that a constant action scores near zero on — the
-strongest candidate is a centre-of-pressure / support-polygon margin term, since the plant
-already carries per-digit touch sensors that make CoP computable; a cheaper stand-in is
-re-enabling `reset_noise_scale` sweeps so the reward has to buy recoveries rather than luck.
-*Touches:* `configs/trex/stage1_balance.toml` `[env]`, plus a new helper in
-`environments/shared/reward_functions.py`.
-*Expected:* the zero-action full-horizon score drops well below the trained score; the
-per-step comparison in F1 inverts.
-*Measurement that decides it:* re-run `zero_action_baseline.py trex` and the F1 per-step
-table. Success is the trained per-step return exceeding the standing statue's, on the same
-seed family. Until that inverts, stage 1 is not measuring balance.
-*Note:* this makes future runs incomparable with §2. Do it once, deliberately, and re-baseline.
+**NS-1 — Change the task, not the reward. (Addresses F1.)**
+
+An earlier draft of this item proposed "add a stance-quality term the statue cannot collect,
+strongest candidate a centre-of-pressure / support-polygon margin." **That was wrong, and the
+reason it is wrong is the most useful thing in this section.**
+
+Four candidate designs were built and each attacked by two independent verifiers who
+implemented the proposed term in a standalone rollout and measured it for the statue and for
+`robust_best_model.zip`. **None survived.** Three of the four add a per-step reward term, and
+all three were refuted 2/2 on the same measured failure — the statue collects the term *at
+least as much as* the trained policy:
+
+| candidate | statue (per 1000 standing steps) | trained policy | winner |
+|---|---|---|---|
+| two-sided height error (repair the existing term) | −1.25, and −0.00 in the settled window | −51.57 | statue by 50.32 |
+| capture-point / support-polygon containment | −392.86 | −1792.58 | statue by 1399.72 |
+| potential-based shaping on divergent CoM velocity | +0.99/ep | −1.45/ep | statue by 2.44 |
+
+Three unrelated mathematical families, same result. That is not three tuning failures, it is
+one structural fact: **the plant is passively stable at the home keyframe, and at any static
+equilibrium the centre of pressure lies exactly under the centre of mass** (measured to 0.2 mm
+over 6126 statue steps — it is forced, since at rest the ground reaction must pass through the
+CoM). Any bounded per-step function of the standing state that means "well balanced" is
+therefore *maximised by standing perfectly still*, which is exactly what `action = 0` does
+under `home-keyframe-residual/v1`. **With no disturbance, the optimal balance controller is the
+statue.** You cannot pay a policy to beat it at a game where it is already the optimum.
+
+Two of the three also made F1 numerically worse: the height repair moves the equal-survival
+deficit 333.66 → 381.32 and drops the achievable stage-1 ceiling to 1850.00 against
+`min_avg_reward = 1840.0`, i.e. it would make the shipped gate near-unclearable and halt the
+curriculum at `train_base.py:1292`.
+
+*Change instead:* apply a scheduled external shove — a runtime write to
+`data.xfrc_applied[root, 0:3]` in `BaseDinoEnv.step`, no reward term, no observation change.
+Every ~2 s, push the root horizontally in a uniformly random direction with an impulse sized at
+1.5× the capture-point velocity (≈150 N for 0.20 s on this plant).
+*Touches:* a new `_apply_perturbation()` in `environments/shared/base_env.py` called at the top
+of `step()`, a pure `external_push_force()` kernel in `environments/shared/reward_functions.py`
+so both backends share it, and new `perturbation_*` keys in `configs/trex/stage1_balance.toml`
+`[env]` defaulting to `0.0` everywhere else.
+
+**It must go in `step()`, not `reset()`.** Verified: `plant_contract.py:916` hashes
+`_callable_semantics(env.reset)` into `policy_interface_revision` for every
+`home-keyframe-residual/v1` species, while `step` appears **zero** times in the interface
+payload. A `step()` hook moves no fingerprint and invalidates no checkpoint.
+
+*Measured, 40 episodes, seed 3042, reproduced independently by two verifiers:*
+
+| | statue | trained checkpoint |
+|---|---|---|
+| shipped (no push) | 1743.73, 57% full-horizon | 2489.65, 100% |
+| push on, noise 0.05 | **711.05 ± 403.76, 0 of 40 full-horizon** | 2418.38 ± 357.61, 85% |
+| push on, noise 0.10 | **604.18 ± 483.99, 0 of 40** | **NOT MEASURED** |
+
+The statue does not merely score less — *the standing statue ceases to exist*, so
+`reward_mean_standing` becomes undefined and the F1 comparison has no left-hand side. Gaming
+was hunted hard and found nothing: 13 hand-designed constant stances, two independent CEM
+searches over the full 21-dim constant action space (best holdouts 490.3 and 692.0, both ≤
+zeros), the trained policy's own settled mean action held constant (184.6), and a blind
+clock-driven brace policy (732.8 mean, worse mean-minus-std, 1 of 40 full-horizon).
+
+*The one load-bearing number that does not exist:* the trained checkpoint has never been
+evaluated at noise 0.10 with the push on. That is a ~10-minute eval, not a training run, and it
+gates everything below.
+
+*Mandatory corrections to the proposal as filed* — each measured, not argued:
+
+1. **Do not bundle `reset_noise_scale` 0.10 → 0.05.** At 0.05 with the push off the statue
+   scores 2568.7 at 90% full-horizon against the checkpoint's 2498.8 — the statue wins
+   outright, reversing the shipped config's +660.5 edge to the policy. Keep 0.10.
+2. **Ship the impulse fixed, not ramped.** `set_reward_weight` (`base_env.py:752`) is a bare
+   `setattr`, so a `RewardRampCallback` on `perturbation_delta_v` would be a step function
+   unless the force is recomputed inside `_apply_perturbation` on every call.
+3. **Jitter the interval** (`perturbation_jitter`). The blind-clock brace exploit is weak but
+   real (+17% mean); one line removes it. Re-measure the floor with jitter on.
+4. **Force `perturbation_delta_v = 0.0` in every diagnostic script.** `zero_action_baseline`,
+   `joint_excursion_report`, `action_bound_report`, `actuator_saturation_report` and
+   `observation_ablation_report` all build the env from the TOML and would silently measure a
+   shoved plant.
+5. **Leave `min_avg_reward = 1840.0` alone** (supersedes NS-2 for T-Rex). The "+5.5% over the
+   measured floor" rule dies with the floor: 1840 is 2.6× the pushed statue and the
+   un-retrained checkpoint clears it by ~580. Rewrite the comment block, not the value.
+
+*Honest scope.* This does **not** repair the per-step reward. Under the push the statue still
+earns more shaping per surviving step than the policy (2.680 vs 2.553); all the separation comes
+from termination plus `fall_penalty`. Stage 1 stays a survival test — it becomes a survival test
+a statue fails. Record that in the TOML rather than claiming the reward now measures balance.
+
+*Worth re-testing afterwards, not before:* the support-polygon term was refuted on an
+**undisturbed** plant, where the statue's capture point barely moves. Once a disturbance exists
+the statue no longer stands and a disturbance-rejection term is no longer maximised by
+stillness. It would still need its two real defects fixed first — a single-support exploit (the
+isotropic `patch_radius²·I` floor lets one loaded digit score containment 1.0000) and the fact
+that 90% of its measured statue deficit is a 13.2 mm touch-site placement artefact rather than a
+stance error.
 
 **NS-2 — Re-derive `min_avg_reward` against the *standing* floor, not the falling one —
 for all four species, not just T-Rex.**
--- a/environments/shared/scripts/zero_action_baseline.py
+++ b/environments/shared/scripts/zero_action_baseline.py
@@ -118,8 +118,16 @@ def score(env, episodes: int, seed: int, noise: float | None = None) -> dict:
         # do-nothing policy survives are worth far more than its average.  A
         # gate set above ``reward_mean`` but below this one is cleared by a
         # policy that has learned only "do not fall".
-        "reward_mean_standing": float(rewards_arr[standing].mean()) if standing.any() else float("nan"),
-        "reward_std_standing": float(rewards_arr[standing].std()) if standing.any() else float("nan"),
+        #
+        # ``None``, not NaN, when the statue never reaches the horizon: this
+        # dict is serialised straight into the baseline records the notebook
+        # writes to Drive, and ``json.dumps`` renders a float NaN as a bare
+        # ``NaN`` literal, which RFC 8259 does not allow.  Python and jq accept
+        # it; strict parsers (JavaScript ``JSON.parse``, most Go/Rust decoders)
+        # reject the whole file.  ``None`` serialises to ``null``.  Brachiosaurus
+        # hits this today -- it never survives a full episode under zero action.
+        "reward_mean_standing": float(rewards_arr[standing].mean()) if standing.any() else None,
+        "reward_std_standing": float(rewards_arr[standing].std()) if standing.any() else None,
         "n_standing": int(standing.sum()),
         "length_mean": float(lengths_arr.mean()),
         "full_horizon_share": float(standing.mean()),
@@ -128,6 +136,13 @@ def score(env, episodes: int, seed: int, noise: float | None = None) -> dict:
         "episodes": int(rewards_arr.size),
         "reset_noise_scale": float(env.reset_noise_scale),
     }
+
+
+def _standing_text(result: dict) -> str:
+    """Format the standing score, which is ``None`` when nothing stood."""
+    if result["n_standing"] == 0:
+        return "-"
+    return f"{result['reward_mean_standing']:.1f}"
 
 
 def report(species: str, stage: int, episodes: int, seed: int, sweep: bool) -> None:
@@ -141,7 +156,7 @@ def report(species: str, stage: int, episodes: int, seed: int, sweep: bool) -> N
                 result = score(env, episodes, seed, noise=noise)
                 print(
                     f"  {noise:>7}{result['reward_mean']:>12.1f}{result['reward_mean_minus_std']:>11.1f}"
-                    f"{result['reward_mean_standing']:>11.1f}"
+                    f"{_standing_text(result):>11}"
                     f"{result['length_mean']:>11.1f}{result['full_horizon_share']:>13.0%}"
                 )
             return
@@ -150,8 +165,13 @@ def report(species: str, stage: int, episodes: int, seed: int, sweep: bool) -> N
         print(f"\n{label}: zero-action baseline ({episodes} episodes, reset_noise={env.reset_noise_scale})")
         print(f"  reward             {result['reward_mean']:.2f} +/- {result['reward_std']:.2f}")
         print(f"  reward mean-std    {result['reward_mean_minus_std']:.2f}")
+        standing = (
+            "never reached the horizon"
+            if result["n_standing"] == 0
+            else f"{result['reward_mean_standing']:.2f} +/- {result['reward_std_standing']:.2f}"
+        )
         print(
-            f"  reward standing    {result['reward_mean_standing']:.2f} +/- {result['reward_std_standing']:.2f}"
+            f"  reward standing    {standing}"
             f"   ({result['n_standing']} of {result['episodes']} episodes reached the horizon)"
         )
         print(f"  episode length     {result['length_mean']:.1f} of {result['horizon']}")
--- a/environments/shared/tests/test_zero_action_baseline.py
+++ b/environments/shared/tests/test_zero_action_baseline.py
@@ -1,0 +1,89 @@
+"""Tests for the zero-action baseline diagnostic.
+
+``score()``'s return value is serialised straight into the baseline records the
+training notebook writes to Google Drive, so it has to survive a strict JSON
+round-trip.  It did not: a species whose statue never reaches the horizon
+produced ``float("nan")``, and ``json.dumps`` renders that as a bare ``NaN``
+literal, which RFC 8259 does not permit.  Brachiosaurus hits that case today.
+"""
+
+import json
+
+import numpy as np
+import pytest
+
+from environments.shared.scripts.zero_action_baseline import score
+
+
+class _StubEnv:
+    """Minimal env stand-in: ``score`` only needs these four members."""
+
+    def __init__(self, lengths, horizon=1000, reset_noise_scale=0.1):
+        self.max_episode_steps = horizon
+        self.reset_noise_scale = reset_noise_scale
+        self.action_space = type("Box", (), {"shape": (3,)})()
+        self._lengths = list(lengths)
+        self._step = 0
+        self._episode = -1
+
+    def reset(self, seed=None):
+        self._episode += 1
+        self._step = 0
+
+    def step(self, action):
+        self._step += 1
+        done = self._step >= self._lengths[self._episode]
+        truncated = done and self._lengths[self._episode] >= self.max_episode_steps
+        info = {} if truncated else {"termination_reason": "fallen"}
+        return None, 1.0, done and not truncated, truncated, info
+
+
+def _strict_json_round_trip(payload):
+    """Serialise and parse with bare NaN/Infinity literals rejected."""
+
+    def _reject(constant):
+        raise ValueError(f"bare {constant} literal is not valid JSON")
+
+    return json.loads(json.dumps(payload), parse_constant=_reject)
+
+
+class TestScoreIsJsonSerialisable:
+    def test_no_standing_episodes_serialises_strictly(self):
+        """The regression: nothing reaches the horizon, so there is no standing score."""
+        result = score(_StubEnv([100, 250, 400]), episodes=3, seed=0)
+
+        assert result["n_standing"] == 0
+        assert result["reward_mean_standing"] is None, (
+            "must be None, not NaN — json.dumps renders NaN as a bare literal that "
+            "strict JSON parsers reject, and this dict is written to Drive"
+        )
+        assert result["reward_std_standing"] is None
+
+        parsed = _strict_json_round_trip(result)
+        assert parsed["reward_mean_standing"] is None
+
+    def test_some_standing_episodes_report_a_number(self):
+        result = score(_StubEnv([100, 1000, 1000]), episodes=3, seed=0)
+
+        assert result["n_standing"] == 2
+        assert result["reward_mean_standing"] == pytest.approx(1000.0)
+        assert result["reward_std_standing"] == pytest.approx(0.0)
+        assert result["full_horizon_share"] == pytest.approx(2 / 3)
+
+        parsed = _strict_json_round_trip(result)
+        assert parsed["reward_mean_standing"] == pytest.approx(1000.0)
+
+    def test_standing_score_exceeds_the_unconditional_mean(self):
+        """The property the number exists for: surviving episodes are worth more."""
+        result = score(_StubEnv([100, 200, 1000, 1000]), episodes=4, seed=0)
+
+        assert result["reward_mean_standing"] > result["reward_mean"], (
+            "a per-step reward makes full-horizon episodes score above the mix, "
+            "which is why a gate calibrated on the unconditional mean under-binds"
+        )
+
+    def test_every_value_is_json_native(self):
+        """No numpy scalars: they serialise inconsistently across versions."""
+        result = score(_StubEnv([100, 1000]), episodes=2, seed=0)
+        for key, value in result.items():
+            assert not isinstance(value, np.generic), f"{key} is a numpy scalar"


kuds:
## Review: the analysis holds and reproduces exactly, but the numbers are stale and the code half is already on `main`

I reproduced this branch's measurements and then re-ran them against current `main`. Summary: **the NS-1 reasoning is correct and I confirmed its numbers to the digit — but every statue figure in the doc predates `435f35f`/`41f950b`, and the reward re-weighting in those commits moved the floor enough to invalidate mandatory correction #5.**

### 1. This branch's numbers reproduce exactly

40 episodes, seed 3042, worktree at `ec23125` (this PR's merge base):

```
reward_mean      1743.7322878268508      (doc: 1743.73 +/- 1275.54)
reward_standing  2834.9516677066094      (doc: 2834.95)
n_standing 23/40   full_horizon 0.575    terminations 23 trunc / 13 nosedive / 4 fallen
```

### 2. But `435f35f` + `41f950b` moved the floor

Same script, same seed, same episode count, at `48fd90a`. I first confirmed `environments/trex/assets/trex.xml` is **byte-identical** between the two commits, so this isolates reward re-weighting with no plant confound:

| | `ec23125` (this PR's base) | `48fd90a` (current `main`) | Δ |
|---|---|---|---|
| `reward_mean` | 1743.7322878268508 | 1971.5736396444001 | **+227.84** |
| `reward_mean_standing` | 2834.9516677066094 | 3244.039005709715 | **+409.09** |
| `length_mean` | 638.15 | 638.15 | **0** |
| `full_horizon_share` | 0.575 | 0.575 | **0** |
| terminations | 23 / 13 / 4 | 23 / 13 / 4 | **0** |

The bottom three rows matter: the statue's trajectories are identical. Nothing about the dynamics changed. The whole +227.84 / +409.09 is the new posture terms paying a do-nothing policy more for doing the same nothing.

Decomposing the statue at the home keyframe (`reset_noise_scale = 0`, settled window of a 1000-step zero-action episode) shows why — it saturates essentially every new term:

```
settled pelvis_z 0.9260   head_tip_z 0.9444   foot forces R 420.5 N  L 420.5 N

term                        weight    value   frac of weight
reward_alive                  1.00   1.0000        1.000
reward_bilateral_support      0.60   0.6000        1.000
reward_height                 0.60   0.6000        1.000
reward_head_clearance         0.35   0.3500        1.000
reward_heading                0.10   0.0998        0.998
reward_leg_home_pose          0.50   0.4903        0.981
reward_neck_posture           0.20   0.1574        0.787
reward_smoothness                —   0.0000    (exactly 0)
reward_energy                    —   0.0000    (exactly 0)
TOTAL                                3.294 /step
```

This is the same structural result §NS-1 already proves — a bounded per-step function of the standing state is maximised by standing still — so it is a confirmation of this PR's argument, not a contradiction of it. Worth noting `support_conditioned_alive_fraction = 0.5` is inert against a statue: `reward_alive` returns full credit because the statue has perfect bilateral support.

### 3. This invalidates mandatory correction #5

Correction #5 argues to leave `min_avg_reward = 1840.0` alone because it is "2.6× the pushed statue" and "the un-retrained checkpoint clears it by ~580." Against the current floor:

- `1840` sat **+96.27 above** the old floor of 1743.73 — it bound, barely.
- `1840` now sits **−131.57 below** the new floor of 1971.57.

The live run `20260729_151044` recorded exactly this in its own pre-flight, verdict `"FAILS — a statue clears this gate"`, and trained anyway because the notebook's pre-flight cell only prints. Every push-on figure in NS-1 (711.05 at noise 0.05, 604.18 at noise 0.10) was measured against the pre-`435f35f` reward and needs re-measuring before the perturbation design is acted on. The already-flagged missing number — checkpoint at noise 0.10 with push on — now needs redoing at current `main` regardless.

### 4. The code changes are already on `main`, and the merged version is a superset

`3a8a37a` ("Fix zero-action baseline JSON output", #472) landed the identical `float("nan")` → `None` change, plus three things this branch lacks:

- **`gate_margin()`** — without it, the notebook's `_gate - _result["reward_mean_standing"]` hits `None` and raises `TypeError`. This branch's fix on its own converts a bad-JSON bug into a notebook crash for brachiosaurus, the exact species cited as the live case.
- **`allow_nan=False`** at both notebook write sites, so a bad write fails loudly.
- A 134-line test at the same path as this branch's 89-line `test_zero_action_baseline.py`.

Both files will conflict on merge.

### Recommendation

Rebase to **doc-only** — drop `environments/shared/scripts/zero_action_baseline.py` and `environments/shared/tests/test_zero_action_baseline.py`, both superseded by `3a8a37a`. Then:

1. Re-measure the NS-1 statue table at `48fd90a` and mark the existing figures with the commit they were taken at.
2. Rewrite correction #5 — its premise is now inverted.
3. Consider retitling. The body reads as an implementation, but the diff contains no `_apply_perturbation`, no `external_push_force`, and no `perturbation_*` keys; it is a recorded design.
4. Split OQ-6 into its own issue. Four touch sensors reading 0.0 N against 1733.3 N of measured contact, with `foot_contact_gate`/`foot_contact_weight` keyed off them, is too live for an open-questions section — and the live T-Rex run shows `unsupported_duty = 0.399` at 4M steps, so the same defect class may not be fully cleared for T-Rex either.

The doc content itself is the strongest part of this PR and should land. Three unrelated candidate families refuted 2/2 by direct measurement, explained by a structural fact about static equilibrium, is a more useful artifact than a fourth tuning attempt would have been — and the honest-scope paragraph is the right disclosure.

<details>
<summary>Repro</summary>

```bash
uv venv .v && uv pip install --python .v/bin/python -e .   # core deps only, no torch/SB3
.v/bin/python environments/shared/scripts/zero_action_baseline.py trex

git worktree add /tmp/pre ec23125
cd /tmp/pre && PYTHONPATH=/tmp/pre ../.v/bin/python \
  environments/shared/scripts/zero_action_baseline.py trex
```

`PYTHONPATH` matters — an editable install silently shadows the worktree and returns current-`main` numbers from both checkouts.
</details>

---

kuds:
## Part 2: live-run evidence, a new defect in `reward_foot_load_balance`, and ranked next steps

Continues [the validation above](https://github.com/kuds/mesozoic-labs/pull/471#issuecomment-5122798715), which covered the reproduction of this branch's numbers, the `ec23125` → `48fd90a` counterfactual, and the duplicate code on `main`. This part adds the live-run measurements, one new defect found while validating, and consolidated recommendations. **No action requested — recording findings.**

---

## A. New defect: `reward_foot_load_balance` pays nothing when airborne

Found while checking whether the term was inert. It isn't inert — it's escapable.

`reward_functions.py` computes `imbalance = |R − L| / (R + L + 1e-8)`. With both feet at zero that is `0 / 1e-8 = 0`, so **no penalty**. Measured directly against the shipped weights (`foot_load_balance_weight = 0.3`, `bilateral_support_weight = 0.6`, `foot_contact_saturation_force = 350`):

| foot forces (N) | imbalance | `load_balance` | `bilateral_support` | **sum** |
|---|---|---|---|---|
| `[420, 420]` both loaded, even | 0.0000 | −0.0000 | 0.6000 | **+0.6000** |
| `[588, 252]` 70/30 split | 0.4000 | −0.1200 | 0.4320 | **+0.3120** |
| `[840, 0]` one foot carries all | 1.0000 | −0.3000 | 0.0000 | **−0.3000** |
| `[0, 0]` **airborne** | **0.0000** | **−0.0000** | 0.0000 | **0.0000** |
| `[1e-4, 0]` near-airborne | 0.9999 | −0.3000 | 0.0000 | **−0.3000** |

Two problems:

1. **Combined, the two foot terms score airborne (0.000) strictly better than single support (−0.300).** On a balance stage that ordering is backwards. `reward_bilateral_support` cannot compensate — it returns 0.0 for both states, so it does not distinguish them.
2. **Knife-edge at zero.** 1e-4 N of contact costs −0.3; exactly 0 N costs nothing. The gradient near the boundary points away from contact.

This is consistent with what the live run actually did. At 4.0M steps `unsupported_duty = 0.399`, `contact_asymmetry` has *risen* 0.681 → 0.854, and `foot_load_imbalance` is pinned at ~0.26 for the entire run. The observed mean penalty is −0.078/step, and −0.3 × 0.26 = −0.078 — the policy is collecting the exemption roughly 40% of the time.

**Proposed fix** — make no-contact the worst case rather than the best:

```python
total = right_force + left_force
imbalance = xp.where(total > min_support_force,
                     xp.abs(right_force - left_force) / (total + 1e-8),
                     1.0)
```

Needs the matching edit on the JAX path in `jax_reward_termination.py` for parity. Neither `test_reward_functions.py` nor `test_trex_mjx_reward_parity.py` currently covers a `[0, 0]` input; both should.

**Caveat that gates this:** if the T-Rex foot touch sensors under-report the way OQ-6 describes for brachiosaurus, then `unsupported_duty = 0.399` is a sensor artefact rather than real airtime, and the severity here changes (the defect in the formula is real either way). See recommendation 1 below.

---

## B. Live run `20260729_151044` — measurements

Commit `48fd90a`, seed 42, 4 envs, L4, started 15:10:44 UTC. State at last read: **4.0M / 6M steps in stage 1**.

**The pre-flight fired and did not stop anything.** `zero_action_baseline.json`, written into the run dir at 15:11:

```
reward_mean 1971.57   standing 3244.04   full-horizon 57.5%   gate 1840
verdict: "FAILS — a statue clears this gate"    margin_over_mean: −131.57
```

Notebook cell 12 prints and saves the verdict but never raises, so training started on a gate it had just measured as cleared by a statue.

**Eval curve** (`evaluations.npz`, 30 episodes, seed 1042):

| steps | reward | ep length |
|---|---|---|
| 1.0M | 166.07 ± 113.30 | 163.9 |
| 2.0M | 695.22 ± 497.01 | 835.2 |
| 3.0M | 1437.54 ± 589.52 | 788.9 |
| 3.5M | 2059.60 ± 361.02 | 965.6 |
| 4.0M | **2215.11 ± 39.63** | **1000.0** |

Under 1000 for the first 3M with real reversals (1.85M: 934.40 → 2.05M: 444.49; 2.6M: 1148.09 → 2.65M: 573.87), stable only past 3.35M, plateaued ~2150–2215 for the last 450k. It cleared `min_avg_reward = 1840` / `min_avg_episode_length = 750` at ~3.5M and has held it since, so it will advance at `train_base.py:1283`.

Optimizer health is fine — this is a reward-shape problem, not a tuning problem: `explained_variance` 0.961, `value_loss` 0.0043, `approx_kl` 0.0120 against a 0.03 target, policy `std` 1.000 → 0.808.

**Stance diagnostics** (`diagnostics.npz`, 489 samples, start → 4.0M):

| metric | start | 4.0M | |
|---|---|---|---|
| `bilateral_support_duty` | 0.084 | 0.471 | improving |
| `unsupported_duty` | 0.692 | 0.399 | improving, still high |
| `single_support_duty` | 0.223 | 0.130 | improving |
| `contact_asymmetry` | 0.681 | **0.854** | **worse** |
| `foot_load_imbalance` | 0.257 | 0.259 | **flat** |
| `leg_home_pose_error` | 0.191 | 0.167 | flat |
| `action_delta` | 21.31 | 12.45 | improving |
| `height_error` | 0.102 | 0.009 | improving |
| `drift_distance` | 0.095 | 0.592 | — |
| `abs_speed` | 0.513 | 0.343 | above the 0.1 penalty threshold |

`bilateral_support` is clearly doing work (0.084 → 0.471, the largest behavioural gain in the batch). `foot_load_balance` and `leg_home_pose` are not moving their own diagnostics.

**Statue vs policy, same commit, same reward function:**

| term | statue (settled) | policy @4.0M | statue edge |
|---|---|---|---|
| `bilateral_support` | 0.600 | 0.252 | **+0.348** |
| `smoothness` | 0.000 | −0.296 | +0.296 |
| `alive` | 1.000 | 0.710 | +0.290 |
| `leg_home_pose` | 0.490 | 0.320 | +0.170 |
| `foot_load_balance` | −0.000 | −0.078 | +0.078 |
| `speed` | −0.000 | −0.073 | +0.073 |
| `energy` | 0.000 | −0.051 | +0.051 |
| `height` | 0.600 | 0.576 | +0.024 |
| `head_clearance` | 0.350 | 0.350 | 0.000 |
| `neck_posture` | 0.157 | 0.183 | −0.026 |
| **total /step** | **3.294** | **1.891** | |

Apples-to-apples (both deterministic, both 1000-step episodes): policy eval **2215.11** vs statue standing **3244.04**.

The policy's largest single deficit is on `bilateral_support` — the flagship new term — and its only win is `neck_posture`. Statue per-term figures are from a `reset_noise_scale = 0` settled window; policy figures are training-rollout means, so per-term rows are indicative while the 2215 vs 3244 comparison is exact.

**Minor:** `reward_strike`, `reward_food`, `reward_proximity`, `reward_claw_proximity` and `reward_gait_symmetry` are `NaN` in all 489 samples of `diagnostics.npz`. Harmless inside an `.npz`, but it is the same "NaN means not-applicable" pattern this PR argues against, and it would become the same bug if any of them is ever summarised into JSON.

---

## C. Observations from completed runs

Stage 1 finals across the last five runs were 1256.87 / 2651.80 / 2503.81 / 2654.66 / 2344.74 — **not comparable to each other or to the current run**, since the reward scale changed on 07-28 and 07-29. Recording that explicitly because the numbers invite comparison.

Two things that predate the stance work and are worth separate attention:

- **Stage 2 exceeds its own velocity cap in every run.** `forward_vel_max = 2.5` (the forward term saturates there), yet measured stage-2 velocity was 2.87 / 2.96 / 6.36 / 4.36 / 8.37 m/s. Speed above the cap earns nothing, so whatever is driving that isn't the term meant to, and the 3× spread across runs is itself unexplained.
- **Stage 3 is regressing.** Final eval 1262.98 → 1241.89 → 1236.06 → 1266.11 → **1126.68**, with mean episode length collapsing 213.5 → 121.6 steps (1.2 s). The most recent run is worst on both.

---

## D. Recommended next steps, ranked

**1. Cross-check the T-Rex foot touch sensors against `mj_contactForce` during a policy rollout.** Do this first — it is a ~10-minute measurement and it determines whether items 2 and 6 are aimed at real numbers. `unsupported_duty = 0.399` on a plant that holds 0.93 m pelvis height and never falls is more consistent with sensor under-reporting than with 40% airtime. Same procedure as OQ-6, but on T-Rex and against a policy rather than the statue.

**2. Fix the airborne hole in `reward_foot_load_balance`** (Part A), with the JAX-path parity edit and `[0, 0]` test cases on both paths.

**3. Make the pre-flight blocking** — `notebooks/sb3_training.ipynb` cell 12, after the verdict loop:

```python
_blocking = {s: r["verdict"] for s, r in results.items()
             if r["verdict"] in ("FAILS — a statue clears this gate",
                                 "CHECK PLANT — the statue never survives a full episode")}
if _blocking and not globals().get("ALLOW_WEAK_GATE", False):
    raise RuntimeError(f"Zero-action pre-flight failed: {_blocking}. "
                       "Set ALLOW_WEAK_GATE = True to override.")
```

Highest value per line here. It would have stopped run `20260729_151044` at 15:11, before 2.5 h of GPU time.

**4. Make the gate a computed margin rather than a TOML constant.** The gate must clear the **standing** floor (3244.04), not the unconditional mean (1971.57) — a policy that has learned only "do not fall" collects the standing rate, which is what the existing `"WEAK — binds only against a falling statue"` verdict is for. Have the pre-flight write the measured floor into the run directory and have the curriculum read `standing_floor × 1.055` instead of a hardcoded number. That retires this whole class of staleness bug.

Note what it implies today: 3244 × 1.055 ≈ 3423 against a policy that tops out at 2215, so the gate becomes unclearable. That is not an argument against computing it — it is F1 stated as a number. Until a task change exists, the honest configuration is `min_avg_reward` unset with pilots run without advancement, which is what the config comment already asks for.

**5. Restore the calibration block deleted in `41f950b`.** ~50 lines documenting the +5.5%-over-floor rule, the `collapse_peak_floor` fallback coupling (`curriculum.py:1014`), and that a stage-1 miss halts the entire curriculum. The value `1840` stayed while its derivation left, which is how the current run happened.

**6. Ablate `leg_home_pose`.** Flat at 0.305 → 0.320 across 4M steps with error barely moving. Set the weight to 0.0, run a 1.5M-step pilot, check whether any stance diagnostic responds. Keep `bilateral_support` — it is clearly earning its place.

**7. Rebase this PR to doc-only** and re-measure its NS-1 statue table at current `main`, per the previous comment. Correction #5's premise is inverted by the new floor.

**8. Split OQ-6 into its own issue** rather than leaving it in an open-questions section.

**Lower priority:** the stage-2 velocity cap and the stage-3 regression above; both predate the stance work and want their own investigation.

---

<details>
<summary>Validation methodology — what was and was not verified</summary>

Environment: `mujoco 3.10.0`, `gymnasium 1.3.0` (matching the run's `provenance.json`), `numpy 2.4.6` (run used 2.0.2). Core deps only, no torch/SB3.

**Directly measured by me:**
- Zero-action baseline at `48fd90a` and at `ec23125`, 40 episodes, seed 3042 — both reproduce their respective recorded values exactly.
- `trex.xml` byte-identical across those commits (`git diff --stat` empty), making the counterfactual reward-only.
- Statue per-term decomposition, `reset_noise_scale = 0`, settled window of a 1000-step zero-action episode.
- `reward_foot_load_balance` / `reward_bilateral_support` evaluated directly on synthetic foot-force vectors.

**Read from the run's own artifacts, not re-measured:** the eval curve (`evaluations.npz`), all stance and algo diagnostics (`diagnostics.npz`), `provenance.json`, `stage_config.json`, `zero_action_baseline.json`, and the completed-run `training_summary.txt` files.

**Not verified:** the trained checkpoint was not re-evaluated (no torch/SB3 available in this environment), so all policy-side numbers come from the run's artifacts. This PR's CI claims (ruff, mypy, 77 trex tests, 1438 shared tests) were not run.
</details>

---

kuds:
## Part 3: corrections, and the completed stage-1 result

Run `20260729_151044` finished stage 1 at 6,004,736 steps and advanced to stage 2. Final data changes two things I asserted in [Part 2](https://github.com/kuds/mesozoic-labs/pull/471#issuecomment-5122863591), both read off a mid-run snapshot. Correcting them here.

---

## Corrections to Part 2

**1. `reward_foot_load_balance` is not inert. That was my error.**

I wrote that it is "a constant tax, not a gradient the policy follows" and that it and `leg_home_pose` are "not moving their own diagnostics." That was measured at 4.0M steps. Across the full 6M it does move:

| | start | @4.0M | @6.0M |
|---|---|---|---|
| `foot_load_imbalance` | 0.257 | 0.259 | **0.212** |
| `reward_foot_load_balance` | −0.077 | −0.078 | **−0.064** |

Small, and it does nothing for the first two-thirds of training, but it is not a flat offset. The claim as I stated it was wrong.

**`leg_home_pose` is genuinely inert** — that half holds, now across all 6M rather than 4M: error 0.191 → 0.167 → 0.168, reward 0.305 → 0.320 → 0.320. Recommendation 6 (ablate it) stands unchanged.

**2. The airborne-exemption severity figure should read 21%, not 40%.**

`unsupported_duty` finished at **0.209**, having been 0.399 at the 4.0M sample I quoted. The defect in the formula is unchanged — I verified `|R − L| / (R + L + 1e-8)` returning 0 for `[0, 0]` numerically against the shipped weights, independent of any run — but the exposure is roughly half what I implied.

**3. Every "@4.0M" figure in Part 2's tables is superseded** by the numbers below.

---

## Stage 1, completed

```
best   2549.69 ± 51.96   at 5,950,000
final  2496.22 ± 78.38   ep_len 1000.0   at 6,000,000
```

Against the statue on the same commit (1971.57 mean / 3244.04 standing):

- **+524.65** over the unconditional mean — double the +244 margin at 4.0M
- **−747.82** below the standing score

The finished policy is materially stronger than the mid-run snapshot suggested. **The structural finding is unchanged: the statue's full-horizon score still exceeds the trained policy's full-horizon score by 30%.**

**Stance diagnostics, full run:**

| metric | start | @4.0M | @6.0M |
|---|---|---|---|
| `bilateral_support_duty` | 0.084 | 0.471 | **0.730** |
| `unsupported_duty` | 0.692 | 0.399 | **0.209** |
| `single_support_duty` | 0.223 | 0.130 | 0.061 |
| `action_delta` | 21.31 | 12.45 | 7.83 |
| `drift_distance` | 0.095 | 0.592 | 0.325 |
| `abs_speed` | 0.513 | 0.343 | 0.204 |
| `foot_load_imbalance` | 0.257 | 0.259 | 0.212 |
| `leg_home_pose_error` | 0.191 | 0.167 | 0.168 |
| `contact_asymmetry` | 0.681 | 0.854 | **0.900** |
| `reward_total` /step | −1.219 | 1.891 | 2.363 |

`bilateral_support_duty` reaching 0.730 and `unsupported_duty` falling to 0.209 is a large, real behavioural gain, and it is attributable to the term added in `435f35f`. Part 2 undersold it.

**New observation:** `contact_asymmetry` is the only metric that degraded monotonically across the entire run — 0.681 → 0.854 → 0.900 — and no reward term targets it directly. Worth its own investigation; it is not explained by `foot_load_imbalance`, which improved over the same window.

---

## Stage 2, in flight

At **901,120 / 8,011,776 steps**. Evals 1114.13 → 1110.75 → 1278.75 → 1382.10 at 50k/250k/500k/750k, every episode full-horizon. `min_avg_reward = 100.0` cleared at the first eval, so `min_avg_forward_vel = 2.0` is the binding gate.

On the velocity concern from Part 2 — **too early to judge, recording so it is not forgotten.** `forward_vel` is 0.401 m/s, max seen 0.471, and 0.0% of samples exceed the 2.5 cap. Prior runs only reached 2.87–8.37 m/s by the end of their 8M budget. Re-check around 4–6M.

Stage-2 stance is drifting back (`bilateral_support_duty` 0.709 → 0.645, `unsupported_duty` 0.220 → 0.259), which is expected for locomotion — walking needs swing phases.

---

## What is unchanged

Nothing in the counterfactual or the gate analysis depends on the mid-run snapshot:

- The floor moved **+227.84 mean / +409.09 standing** with byte-identical statue trajectories — reward re-weighting only.
- `min_avg_reward = 1840` went from **+96.27 above** the floor to **−131.57 below** it, inverting mandatory correction #5.
- The pre-flight recorded `"FAILS — a statue clears this gate"` at 15:11 and training proceeded, because the notebook cell only prints.
- The airborne hole in `reward_foot_load_balance` (formula-level, verified independently of the run).
- `3a8a37a` supersedes this branch's code changes, and carries `gate_margin()`, which this branch lacks.

---

## Revised bottom line

The posture work is **better than I judged in Part 2**, and I want that on the record as clearly as the criticism. Killing the any-foot bonus and replacing it with `min`-over-feet bilateral support took `bilateral_support_duty` from 0.084 to 0.730 and `unsupported_duty` from 0.692 to 0.209. That is the stage doing what it is named for, and the stance diagnostics added in `75bcd80` are what made it visible.

The F1 conclusion is unchanged and does not conflict with that. Six bounded per-step terms, all saturated by a statue at the home keyframe, raised the do-nothing floor by more than they raised the policy — measured, with zero behavioural confound. Posture shaping controls what the stance *looks like*; only a task change controls whether the stage *measures balance*. Both things are true at once, and this PR's NS-1 argument is the correct treatment of the second.

Recommendation ranking from Part 2 stands, with one demotion: **item 6 (ablate) should now target `leg_home_pose` only**, not `foot_load_balance`, which earns a small but real keep pending the formula fix in item 2.

---

kuds:
## Part 4: the stage-1 policy is bouncing, not standing — and the gait metric cannot see it

Prompted by an observation that the rendered stage-1 policy appears to take many small steps. It does, and the run's own diagnostics explain it. This also **corrects a claim I made in [Part 3](https://github.com/kuds/mesozoic-labs/pull/471#issuecomment-5123344950)**.

---

## A. The duty split is a hop, not a stance

Final stage-1 values:

| metric | start | @6.0M |
|---|---|---|
| `bilateral_support_duty` | 0.084 | **0.730** |
| `single_support_duty` | 0.223 | **0.061** |
| `unsupported_duty` | 0.692 | **0.209** |

Walking requires sustained **single** support — one foot loaded while the other swings. This policy spends 73% with both feet down, 21% with **neither** foot down, and 6% on one foot. It cycles between "both down" and "both up": a two-footed hop in place.

Training drove it there on purpose. `single_support_duty` fell **0.223 → 0.061**, a 3.6× reduction, while `unsupported_duty` stayed substantial.

## B. The reward makes single support the worst available state

Measured against the shipped weights (`bilateral_support_weight = 0.6`, `foot_load_balance_weight = 0.3`, `foot_contact_saturation_force = 350`):

| state | `bilateral_support` | `foot_load_balance` | **sum** |
|---|---|---|---|
| both feet down, even | +0.600 | −0.000 | **+0.600** |
| **airborne** | 0.000 | **−0.000** | **0.000** |
| one foot carries load | 0.000 | **−0.300** | **−0.300** |

To correct its balance the policy can shift weight onto one leg (−0.300/step) or leave the ground (0.000/step). Leaving the ground is cheaper, so it bounces.

**This elevates the `reward_foot_load_balance` airborne hole from Part 2.** It is no longer only a bad value at `[0, 0]` — it is selecting the gait. Recommendation 2 (the `xp.where(total > min_support_force, …, 1.0)` fix) should move to the front of the queue, ahead of everything except the sensor cross-check that validates the inputs.

## C. Correction to Part 3: `contact_asymmetry`

I wrote that `contact_asymmetry` "degraded monotonically across the entire run — 0.681 → 0.854 → 0.900." **That reading was wrong in direction.** The field is `alternation_ratio` under a legacy alias (`trex_env.py:670`), and by its own definition 1.0 = perfectly alternating, so the rise is nominally an improvement.

It does not matter, because the metric cannot measure what its name implies. Running the real `_compute_gait_symmetry` against synthetic contact sequences:

```
synchronized BOUNCE (both feet together)   alternation_ratio = 1.000
alternating WALK (true stepping)           alternation_ratio = 1.000
STATUE (never breaks contact)              alternation_ratio = 1.000
LIMP (R,R,L repeating)                     alternation_ratio = 0.684
```

Root cause at `base_env.py:512-515` — on a simultaneous two-foot landing the sequence appends `"R"` then `"L"`, so every synchronized bounce records as a textbook alternation:

```python
if r_touchdown:
    self._touchdown_sequence.append("R")
if l_touchdown:
    self._touchdown_sequence.append("L")
```

A statue scores 1.000 too: both feet register a touchdown on the first contact, giving `[R, L]`, and the ratio never updates again. `_compute_stride_frequency` (`metrics.py:384`) shares the blind spot — it counts onsets of *combined* contact, so it would report the bounce rate as a stride frequency.

So the rise 0.681 → 0.900 most plausibly reflects the bounce becoming more regular, not the gait improving. Both my original claim and the metric are unreliable; the duty split in section A is the trustworthy signal.

---

## D. Recommended stage-1 diagnostics

The through-line in this PR is that stage 1 has no instrument that separates *standing* from *not yet fallen*, and now also none that separates *standing* from *hopping*. Six additions, ordered by value per unit of work:

**1. Report every stage-1 metric as a margin over the zero-action baseline.** The pre-flight already computes the statue's score. Carry it into the eval plots as a horizontal line and into `training_summary.txt` as a column. A run whose curve sits below the statue line should be impossible to miss. This alone would have surfaced F1 in July without anyone going looking.

**2. Ground-reaction-force sanity check against body weight.** For any periodic motion the time-averaged total GRF must equal body weight — a physics invariant, so any deviation is a sensor or accounting bug. Log `mean(total_contact_force) / (m·g)` and alarm when it leaves roughly `[0.95, 1.05]`.

This would already be firing. The statue's static total is 420.5 + 420.5 = **841 N**, which is body weight by definition. The policy's logged mean total is 712.6 + 747.8 = **1460 N**, or 1.74× — impossible for a body in periodic motion. I have not isolated whether that is per-toe sensors double-counting into the pad reading (`mjx_env.py:700-702`), averaging that excludes airborne samples, or a genuine sensor fault, but it is the same defect family as OQ-6 and this check turns that investigation into a permanent automated guard for all four species.

**3. Support-state transition matrix.** Log counts between `{bilateral, single-L, single-R, airborne}`. A walk is dominated by `single-L ↔ single-R`; a bounce by `bilateral ↔ airborne`. One 4×4 matrix identifies the gait mode unambiguously. The cheap scalar version is `single_support_duty / unsupported_duty` — 0.29 here, and anything below ~1.0 on a *balance* stage should be a red flag.

**4. Centre-of-pressure diagnostics.** CoP position relative to the support polygon, plus CoP excursion and velocity. This is the actual biomechanical definition of balance, and it separates the three cases the current metrics conflate: a statue has CoP dead centre with zero excursion, a balancing policy makes small bounded corrections, and a bouncing policy has no CoP at all for 21% of the episode. Given NS-1's argument rests on where the CoP sits relative to the CoM, having it as a logged observable would put that reasoning on measured ground.

**5. Vertical oscillation and flight-phase count.** Log `std(pelvis_z)` and peak-to-peak within an episode alongside the mean, plus flight phases per episode and mean bout duration. The mean pelvis height here is 0.932 against the statue's 0.926 — essentially identical, so the mean hides the entire behaviour. The variance is the signal.

**6. Fix `alternation_ratio`, or stop reporting it.** Record a simultaneous two-foot touchdown as one `BOTH` event rather than `R` then `L`. A bounce then scores 0.0 and a walk 1.0, which is the intended semantics. Until then it is worse than no metric, because it reads 1.000 for the failure mode. The legacy `contact_asymmetry` alias should also go — it inverts the meaning of the name.

**Minor:** `reward_strike`, `reward_food`, `reward_proximity`, `reward_claw_proximity` and `reward_gait_symmetry` are `NaN` in all 489 samples of the stage-1 `diagnostics.npz`, and the `term_*` arrays have a different length from the rest, which makes them awkward to index alongside. Both are small cleanups in the same area this PR touches.

---

<details>
<summary>Method</summary>

Duty and reward-component figures are read from the run's own `stage1/diagnostics.npz` (733 samples, final timestep 6,004,736). The state-cost table and the `alternation_ratio` results were produced by calling the repository's own `reward_bilateral_support`, `reward_foot_load_balance` and `_compute_gait_symmetry` directly on synthetic inputs.

I attempted a deterministic rollout of `stage1/models/robust_best_model.zip` to add a bounce frequency in Hz and airborne bout durations. The `VecNormalize` observation statistics could not be transferred into this environment intact, so that number is missing; nothing above depends on it.
</details>

---

kuds:
## Part 5: all four species' stage-1 gates are cleared by a statue

The gate problem is not trex-specific. Measured just now at `48fd90a`, 40 episodes, seed 3042 — the same protocol the notebook pre-flight uses:

```
species           statue mean  standing  full-hz    gate  verdict
trex                   1971.6    3244.0     57%    1840  FAILS - a statue clears this gate
velociraptor           1704.9    1746.4     98%     100  FAILS - a statue clears this gate
brachiosaurus           108.2         -      0%     100  FAILS - a statue clears this gate
dibothrosuchus         1702.0    2594.4     65%     100  FAILS - a statue clears this gate
```

**Every stage-1 curriculum gate in the repository is currently cleared by a do-nothing policy.**

Three of the four still sit at `min_avg_reward = 100.0`, the placeholder value the trex config comment history described as one a statue "cleared seventeen times over." The velociraptor clears its gate by exactly **17.0×**, and its statue survives 98% of full-length episodes — a passively stable plant scoring 1704.9 against a threshold of 100.

**Brachiosaurus fails twice.** Its statue clears the gate *and* never survives a full episode: `full_horizon_share = 0%`, `n_standing = 0`. A plant that falls over under its own home controller is not ready for a balance stage, which is the notebook's `"CHECK PLANT — the statue never survives a full episode"` verdict; it is masked here only because the `FAILS` branch is evaluated first. This is also a live reproduction of the `reward_mean_standing = None` case this PR's NaN fix exists to handle — the `-` in the standing column above is that `None`.

## Why this happened, and what fixes the class

The trex gate was derived correctly: measure the statue (1743.73), add 5.5%. Good methodology, and the ~50-line comment documenting it was careful work. **The failure is that the number outlived its derivation.** `435f35f` changed the reward, the floor moved to 1971.57, and 1840 went from just above the floor to below it. `41f950b` then deleted the comment explaining where 1840 came from while leaving 1840 in place.

The other three were never derived at all — they are still at the default.

This is what recommendation 4 in [Part 2](https://github.com/kuds/mesozoic-labs/pull/471#issuecomment-5122863591) was aimed at, and the four-species result is the argument for it. The pre-flight already measures the floor on every run and writes `zero_action_baseline.json` into the run directory. Have the curriculum read that run's own measured floor and apply the margin — `gate = standing_floor × 1.055` — rather than a TOML literal. Same rule, evaluated fresh. It cannot go stale because there is no stored number to go stale.

Two refinements:

- **Use the standing floor, not the mean.** A policy that has learned only "do not fall" collects the standing rate, so that is what it must beat. The notebook's verdict ladder already encodes this with its `"WEAK — binds only against a falling statue"` tier.
- **The honest consequence:** trex standing is 3244.04, ×1.055 ≈ 3423, against a finished stage-1 policy at 2496.22. The computed gate fails today. That is not an argument against computing it — it is F1 stated as a number. Until a task change exists, the honest configuration is `min_avg_reward` unset with pilots run without curriculum advancement, which is what the trex config comment already asks for.

**If only one change is made, make the pre-flight blocking** (recommendation 3). Five lines, no refactor, and it catches staleness the moment it appears — for today's trex run and for all three other species the next time they train. The computed gate is the thorough version; the blocking pre-flight captures most of the value for a fraction of the work.

<details>
<summary>Repro</summary>

```python
from environments.shared.scripts.zero_action_baseline import build_env, score
from environments.shared.config import load_stage_config

for sp in ["trex", "velociraptor", "brachiosaurus", "dibothrosuchus"]:
    env = build_env(sp, 1)
    try:
        r = score(env, 40, 3042)
    finally:
        env.close()
    gate = load_stage_config(sp, 1)["curriculum_kwargs"].get("min_avg_reward")
    print(sp, r["reward_mean"], r["reward_mean_standing"], r["full_horizon_share"], gate)
```

Core deps only (`mujoco==3.10.0`, `gymnasium 1.3.0`) — no torch or SB3 needed. Roughly four minutes for all four species. The notebook's pre-flight cell already supports this directly via `BASELINE_SPECIES = ["velociraptor", "brachiosaurus", "dibothrosuchus", "trex"]`.
</details>

---

kuds:
# Consolidated review — supersedes my comments 1–5

This replaces [1](https://github.com/kuds/mesozoic-labs/pull/471#issuecomment-5122798715), [2](https://github.com/kuds/mesozoic-labs/pull/471#issuecomment-5122863591), [3](https://github.com/kuds/mesozoic-labs/pull/471#issuecomment-5123344950), [4](https://github.com/kuds/mesozoic-labs/pull/471#issuecomment-5123520329) and [5](https://github.com/kuds/mesozoic-labs/pull/471#issuecomment-5123608512), which were written incrementally against a run that was still in flight and contain two claims the final data contradicts. Read this one instead; §7 lists what I got wrong.

**Verdict on the PR: the doc content should land, the code should be dropped as superseded, and the NS-1 figures need re-measuring before the design is acted on.**

---

## 1. This branch's measurements reproduce exactly

40 episodes, seed 3042, worktree at `ec23125` (this PR's merge base):

```
reward_mean      1743.7322878268508      doc: 1743.73 +/- 1275.54
reward_standing  2834.9516677066094      doc: 2834.95
n_standing 23/40   full_horizon 0.575    terminations 23 trunc / 13 nosedive / 4 fallen
```

The NS-1 argument is sound and its numbers are real. Everything below is about what changed after they were taken.

## 2. The stance commits moved the floor — reward-only, no behavioural confound

`environments/trex/assets/trex.xml` is **byte-identical** between `ec23125` and `48fd90a`, so this isolates the reward re-weighting:

| | `ec23125` | `48fd90a` | Δ |
|---|---|---|---|
| `reward_mean` | 1743.73 | 1971.57 | **+227.84** |
| `reward_mean_standing` | 2834.95 | 3244.04 | **+409.09** |
| `length_mean` | 638.15 | 638.15 | **0** |
| `full_horizon_share` | 0.575 | 0.575 | **0** |
| terminations | 23 / 13 / 4 | 23 / 13 / 4 | **0** |

The statue's trajectories are identical. The entire gain is the new posture terms paying a do-nothing policy more for doing the same nothing.

Decomposing the statue at the home keyframe (`reset_noise_scale = 0`, settled window of a 1000-step zero-action episode) shows why — it saturates nearly every new term:

```
settled pelvis_z 0.9260   head_tip_z 0.9444   foot forces R 420.5 N  L 420.5 N

term                        weight    value   frac of weight
reward_alive                  1.00   1.0000        1.000
reward_bilateral_support      0.60   0.6000        1.000
reward_height                 0.60   0.6000        1.000
reward_head_clearance         0.35   0.3500        1.000
reward_heading                0.10   0.0998        0.998
reward_leg_home_pose          0.50   0.4903        0.981
reward_neck_posture           0.20   0.1574        0.787
reward_smoothness                —   0.0000    (exactly 0)
reward_energy                    —   0.0000    (exactly 0)
TOTAL                                3.294 /step
```

This confirms §NS-1's structural claim rather than contradicting it: a bounded per-step function of the standing state is maximised by standing still. `support_conditioned_alive_fraction = 0.5` is inert against a statue — `reward_alive` returns full credit because the statue has perfect bilateral support.

## 3. Every stage-1 gate in the repository is cleared by a statue

Measured at `48fd90a`, 40 episodes, seed 3042:

```
species           statue mean  standing  full-hz    gate  verdict
trex                   1971.6    3244.0     57%    1840  FAILS - a statue clears this gate
velociraptor           1704.9    1746.4     98%     100  FAILS - a statue clears this gate
brachiosaurus           108.2         -      0%     100  FAILS - a statue clears this gate
dibothrosuchus         1702.0    2594.4     65%     100  FAILS - a statue clears this gate
```

Three still sit at the `100.0` placeholder. The velociraptor clears its gate by exactly **17.0×**, with a statue that survives 98% of full-length episodes.

**Brachiosaurus fails twice** — its statue clears the gate *and* never survives a full episode (`full_horizon_share = 0%`, `n_standing = 0`). That is the `"CHECK PLANT"` condition, masked because `FAILS` is evaluated first. It is also a live reproduction of the `reward_mean_standing = None` case this PR's NaN fix exists for: the `-` in the standing column is that `None`.

**This invalidates mandatory correction #5.** It argues for keeping `1840.0` because it is "2.6× the pushed statue." Against the current floor, `1840` moved from **+96.27 above** to **−131.57 below**. Every push-on figure in NS-1 (711.05 at noise 0.05, 604.18 at noise 0.10) predates `435f35f` and needs re-measuring.

Run `20260729_151044` recorded `"FAILS — a statue clears this gate"` in its own pre-flight at 15:11 and trained anyway, because the notebook cell only prints.

## 4. Stage 1 completed — the policy is bouncing, not standing

```
best   2549.69 ± 51.96   at 5,950,000
final  2496.22 ± 78.38   ep_len 1000.0   at 6,000,000
   +524.65 over the statue's mean (1971.57)
   -747.82 below the statue's standing score (3244.04)
```

| metric | start | @6.0M |
|---|---|---|
| `bilateral_support_duty` | 0.084 | **0.730** |
| `single_support_duty` | 0.223 | **0.061** |
| `unsupported_duty` | 0.692 | **0.209** |
| `action_delta` | 21.31 | 7.83 |
| `foot_load_imbalance` | 0.257 | 0.212 |
| `leg_home_pose_error` | 0.191 | 0.168 |
| `reward_total` /step | −1.219 | 2.363 |

Walking requires sustained **single** support. This policy spends 73% with both feet down, 21% with **neither** foot down, and 6% on one foot — a two-footed hop in place. Training drove `single_support_duty` down 3.6× (0.223 → 0.061).

**The reward selects that gait**, measured against the shipped weights:

| state | `bilateral_support` | `foot_load_balance` | **sum** |
|---|---|---|---|
| both feet down, even | +0.600 | −0.000 | **+0.600** |
| **airborne** | 0.000 | **−0.000** | **0.000** |
| one foot carries load | 0.000 | **−0.300** | **−0.300** |

`reward_foot_load_balance` computes `|R − L| / (R + L + 1e-8)`, which returns 0 when both feet read zero. Single support is therefore strictly the worst of the three states, so a policy needing to correct its balance leaves the ground rather than shift weight.

**The gait metric cannot detect this.** `contact_asymmetry` is `alternation_ratio` under a legacy alias (`trex_env.py:670`). Running the real `_compute_gait_symmetry` on synthetic sequences:

```
synchronized BOUNCE (both feet together)   1.000
alternating WALK (true stepping)           1.000
STATUE (never breaks contact)              1.000
LIMP (R,R,L repeating)                     0.684
```

Root cause at `base_env.py:512-515` — a simultaneous two-foot landing appends `"R"` then `"L"`, so every bounce registers as a textbook alternation. `_compute_stride_frequency` (`metrics.py:384`) shares the blind spot: it counts onsets of *combined* contact, so it would report the bounce rate as stride frequency.

## 5. The code in this PR is superseded

`3a8a37a` (#472) landed the identical `float("nan")` → `None` change plus three things this branch lacks:

- **`gate_margin()`** — without it the notebook's `_gate - _result["reward_mean_standing"]` hits `None` and raises `TypeError`. This branch's fix alone converts a bad-JSON bug into a notebook crash for brachiosaurus — which §3 shows is the live case *today*.
- **`allow_nan=False`** at both notebook write sites.
- A 134-line test at the same path as this branch's 89-line one.

Both files will conflict on merge.

## 6. Recommendations, ranked

1. **Cross-check the T-Rex foot touch sensors against `mj_contactForce` during a policy rollout.** Do this first — it validates the inputs everything else rests on. Motivation in §8.
2. **Fix the airborne hole in `reward_foot_load_balance`** — `xp.where(total > min_support_force, |R−L|/(total+1e-8), 1.0)`, with the JAX-path parity edit and `[0, 0]` cases in `test_reward_functions.py` and `test_trex_mjx_reward_parity.py`. Neither covers it today. This is selecting the gait, not just returning a bad value.
3. **Make the pre-flight blocking** — `notebooks/sb3_training.ipynb` cell 12:

```python
_blocking = {s: r["verdict"] for s, r in results.items()
             if r["verdict"] in ("FAILS — a statue clears this gate",
                                 "CHECK PLANT — the statue never survives a full episode")}
if _blocking and not globals().get("ALLOW_WEAK_GATE", False):
    raise RuntimeError(f"Zero-action pre-flight failed: {_blocking}. "
                       "Set ALLOW_WEAK_GATE = True to override.")
```

   Five lines, no refactor, and it catches all four species. Highest value per line here.
4. **Resolve `min_avg_reward` from the measured floor instead of a TOML literal** — specified in §9.
5. **Restore the calibration block deleted in `41f950b`**, including the `collapse_peak_floor` coupling at `curriculum.py:1014` and the note that a stage-1 miss halts the curriculum.
6. **Ablate `leg_home_pose`** — flat across all 6M (error 0.191 → 0.167 → 0.168, reward 0.305 → 0.320 → 0.320). Keep `bilateral_support`, which is doing the most work of anything in the batch.
7. **Rebase this PR to doc-only** and re-measure the NS-1 statue table at current `main`; rewrite correction #5.
8. **Split OQ-6 into its own issue.**

Lower priority: stage 2's `forward_vel_max = 2.5` against measured 2.87–8.37 m/s in prior runs (the current run is at 0.401 m/s at 901k steps, so re-check at 4–6M); and the stage-3 regression, 1266.11 → 1126.68 with episode length 208.7 → 121.6 steps.

## 7. Corrections to my earlier comments

- **`reward_foot_load_balance` is not inert.** I called it "a constant tax, not a gradient the policy follows" from the 4.0M snapshot. Over the full run `foot_load_imbalance` goes 0.257 → 0.212 and the penalty −0.077 → −0.064. Small and late, but real. `leg_home_pose` *is* inert; that half holds.
- **`contact_asymmetry` did not "degrade monotonically."** I had the direction backwards — 1.0 is nominally *better* by the metric's own definition. It is unreliable either way, per §4.
- **The airborne exposure is 21%, not 40%.** `unsupported_duty` finished at 0.209, having been 0.399 at the 4.0M sample I quoted.
- Every `@4.0M` figure in my earlier comments is superseded by §4.

## 8. Diagnostics worth adding to stage 1

Stage 1 currently has no instrument separating *standing* from *not yet fallen*, and none separating *standing* from *hopping*.

1. **Report every stage-1 metric as a margin over the zero-action floor** — the pre-flight already computes it; put it on the eval plot as a horizontal line and in `training_summary.txt` as a column.
2. **Ground-reaction-force check against body weight.** Time-averaged total GRF must equal body weight for any periodic motion — a physics invariant, so deviation is a sensor or accounting bug. Log `mean(total_contact_force) / (m·g)`, alarm outside ~`[0.95, 1.05]`. **This would already be firing:** the statue's static total is 841 N (body weight by definition) while the policy's logged mean is 1460 N, 1.74×, which is impossible for periodic motion. I have not isolated whether that is per-toe sensors double-counting into the pad reading (`mjx_env.py:700-702`), averaging that excludes airborne samples, or a genuine fault — but it is the same family as OQ-6, and this turns that investigation into a permanent automated guard for all four species.
3. **Support-state transition matrix** over `{bilateral, single-L, single-R, airborne}`. A walk is dominated by `single-L ↔ single-R`, a bounce by `bilateral ↔ airborne`. Cheap scalar: `single_support_duty / unsupported_duty`, which is 0.29 here.
4. **Centre-of-pressure position, excursion and velocity** — the actual biomechanical definition of balance, and the observable NS-1's argument rests on.
5. **Vertical oscillation and flight-phase count** — `std(pelvis_z)` and peak-to-peak alongside the mean. Mean pelvis height is 0.932 against the statue's 0.926, so the mean hides the entire behaviour.
6. **Fix `alternation_ratio` or stop reporting it** — record a simultaneous two-foot touchdown as one `BOTH` event. Today it reads 1.000 for the failure mode.
7. **Per-term reward delta against the statue**, from the pre-flight's own breakdown.
8. **Action power spectrum** — `action_delta` is a scalar mean and cannot distinguish smooth motion from high-frequency chatter.

Minor: `reward_strike`, `reward_food`, `reward_proximity`, `reward_claw_proximity` and `reward_gait_symmetry` are `NaN` in all 489 stage-1 diagnostic samples, and the `term_*` arrays have a different length from the rest.

## 9. What a computed gate looks like

The trex gate was derived correctly — measure the statue, add 5.5%. **The failure is that the number outlived its derivation**: `435f35f` moved the floor, and `41f950b` deleted the comment explaining where `1840` came from while leaving `1840`. The other three species were never derived at all.

`min_avg_reward` is read in a dozen places (`curriculum.py:325`, `species_catalog.py:367`, `jax_curriculum.py:35`, `reporting.py:90`, `visualization.py:114`, `eval_diagnostics.py:247`, `result_bundle.py:850`, the sweep code). So keep it a resolved float everywhere downstream and change only *where the float comes from* — resolve once at run start, freeze for the run.

Config declares the rule rather than the answer:

```toml
[curriculum]
timesteps = 6000000
min_avg_episode_length = 750
gate_reference     = "standing"   # "standing" | "mean" | "none"
gate_margin_fraction = 0.055      # the existing +5.5%-over-floor rule
min_avg_reward     = 1840.0       # fallback only, when no baseline is available
collapse_peak_floor = 2200.0      # set explicitly - see below
```

```python
def resolve_reward_gate(curriculum: dict, baseline: dict | None) -> tuple[float | None, dict]:
    """Resolve min_avg_reward from a measured zero-action baseline.

    Returns (gate, provenance).  Falls back to the config literal whenever no
    measured floor exists, so offline tooling, sweeps and tests keep working.
    """
    reference = curriculum.get("gate_reference", "none")
    literal = curriculum.get("min_avg_reward")
    if reference == "none" or baseline is None:
        return literal, {"source": "config_literal"}

    floor = baseline["reward_mean_standing"] if reference == "standing" else baseline["reward_mean"]
    if floor is None:
        # The statue never reaches the horizon.  A plant that falls over under
        # its own home controller is not ready for a balance stage at all.
        return literal, {"source": "config_literal", "reason": "no_standing_floor",
                         "action_required": "CHECK PLANT"}

    margin = float(curriculum.get("gate_margin_fraction", 0.055))
    return floor * (1.0 + margin), {
        "source": "measured",
        "reference": reference,
        "floor": floor,
        "margin_fraction": margin,
        "baseline_captured_at": baseline["captured_at"],
        "plant_identity": baseline["plant_identity"],
    }
```

Five design points:

**Gate on the standing floor, not the mean.** A policy that has learned only "do not fall" collects the standing rate. The notebook's verdict ladder already encodes this with its `"WEAK — binds only against a falling statue"` tier. There is also a statistical reason: the statue's reward distribution is bimodal (fell vs stood), so the mean sits between two modes and is unstable — trex `reward_std` is 1486.52 against `reward_std_standing` of just 23.55. The standing mode is tight and well-behaved; the mean is not. That argues for eventually replacing `× 1.055` with `standing_mean + k · std_standing`, which is scale-free.

**Persist the provenance, not just the number.** Write the resolved gate and its derivation into the run directory next to `zero_action_baseline.json`. That is precisely what was lost when the comment was deleted — the run's artifacts should carry the reasoning.

**Resolve once per run, then freeze.** Not per-eval; the comparison must be stable and reproducible within a run.

**Decouple the collapse detector.** `curriculum.py:1014` falls back to `min_avg_reward` for `peak_floor` when `collapse_peak_floor` is unset, and no config sets it. If the resolved gate jumps to ~3423, collapse detection arms only above 3423 — i.e. never. This coupling was documented in the block `41f950b` deleted and would fail silently. Set `collapse_peak_floor` explicitly wherever this lands.

**Expect it to fail today, and make that a first-class state.** Applying `standing × 1.055` right now:

| species | resolved gate | current best |
|---|---|---|
| trex | 3422.5 | 2496.22 |
| velociraptor | 1842.5 | — |
| brachiosaurus | `CHECK PLANT` | — |
| dibothrosuchus | 2737.1 | — |

The trex gate becomes unclearable. That is not a flaw in the mechanism — it is F1 expressed as a number instead of a paragraph. So the resolver needs an explicit `advancement = "disabled"` pilot mode where the gate is resolved and reported but does not halt the curriculum. That is exactly what the current trex config comment asks for in prose; this makes it a setting rather than a note someone has to remember.

**If only one thing ships, ship recommendation 3 instead.** The blocking pre-flight is five lines and catches the same staleness the moment it appears, for all four species. §9 is the thorough version, not the urgent one.

---

<details>
<summary>Method — what was and was not verified</summary>

`mujoco 3.10.0`, `gymnasium 1.3.0` (matching the run's `provenance.json`), `numpy 2.4.6` (run used 2.0.2).

**Directly measured:** zero-action baselines for all four species at `48fd90a` and for trex at `ec23125`, 40 episodes, seed 3042; `trex.xml` byte-identity across those commits; the statue's per-term decomposition at `reset_noise_scale = 0`; `reward_bilateral_support`, `reward_foot_load_balance` and `_compute_gait_symmetry` evaluated directly on synthetic inputs.

**Read from the run's artifacts, not re-measured:** eval curves and all stance/algo diagnostics for stages 1 and 2, `provenance.json`, `stage_config.json`, `zero_action_baseline.json`, and the completed-run `training_summary.txt` files.

**Not verified:** the trained checkpoint was not re-rolled out — I could not transfer the `VecNormalize` observation statistics into this environment intact, so there is no bounce frequency in Hz or airborne bout duration. Nothing above depends on it. This PR's CI claims (ruff, mypy, 77 trex tests, 1438 shared tests) were not run.

**Repro for §3:**

```python
from environments.shared.scripts.zero_action_baseline import build_env, score
from environments.shared.config import load_stage_config

for sp in ["trex", "velociraptor", "brachiosaurus", "dibothrosuchus"]:
    env = build_env(sp, 1)
    try:
        r = score(env, 40, 3042)
    finally:
        env.close()
    gate = load_stage_config(sp, 1)["curriculum_kwargs"].get("min_avg_reward")
    print(sp, r["reward_mean"], r["reward_mean_standing"], r["full_horizon_share"], gate)
```

Core deps only, ~4 minutes for all four. For the `ec23125` comparison, `PYTHONPATH` must point at the worktree — an editable install silently shadows it and returns current-`main` numbers from both checkouts.
</details>

---

kuds:
## Independent adversarial follow-up: keep the computed-gate lifecycle, replace the standing-floor formula

I re-checked the PR description, diff, all six existing comments, the PR branch/base, current `main`, every major `min_avg_reward` consumer, the available run artifacts, and a compatible existing checkpoint.

**Bottom line:** resolving a baseline-calibrated gate once at run creation, freezing it, and persisting its provenance is a sound design premise. The proposed calculation

```text
reward_mean_standing × 1.055
```

is not a sound advancement gate. It compares different quantities:

- the trained policy is gated on **unconditional** mean episode return;
- `reward_mean_standing` is the statue's return **conditioned on full-horizon survival**.

Conditioning removes the zero-action failure mode that a trained controller is supposed to improve.

### Direct validation

Snapshots:

- PR base recorded by GitHub: `ec2312543718ee618e0327292b5edc93b36edf6e`
- PR head: `175542a8ae9ae6984435f7e3ad0cb84b24a046a5`
- current `main`: `48fd90afc1bef16f7f45054f31571cd99dfc6fb1`

Current-main T-Rex zero action, three disjoint 40-seed blocks:

| seeds | unconditional mean | standing mean | standing n | mean length |
|---|---:|---:|---:|---:|
| 3042–3081 | 1971.57 | 3244.04 | 23/40 | 638.15 |
| 4042–4081 | 1968.72 | 3250.45 | 23/40 | 638.45 |
| 5042–5081 | 1884.71 | 3233.99 | 22/40 | 615.40 |

I recovered the existing run `20260729_151044` checkpoint and matched VecNormalize artifact:

```text
model sha256:   ccb6f322df91215c2aab2aa41e81a3c33bf3e3870bdecad7f64539214b950549
vecnorm sha256: 3a6367903b639eba8d8a2c772c2b58306dcb142963699b903e689e232e9bc9e1
```

I then evaluated policy and zero action with **exactly paired per-episode seeds**:

| seeds | policy mean | zero mean | standing statue | paired mean improvement | approximate one-sided 95% t LCB |
|---|---:|---:|---:|---:|---:|
| 3042–3081 | 2475.48 | 1971.57 | 3244.04 | +503.91 | +114.69 |
| 4042–4081 | 2496.52 | 1968.72 | 3250.45 | +527.80 | +133.26 |
| 5042–5081 | 2557.05 | 1884.71 | 3233.99 | +672.34 | +269.74 |

Across 120 paired seeds, the policy averaged `2509.69`, zero action averaged `1941.67`, and the paired improvement averaged `+568.02`; the policy reached the horizon 118/120 times versus 68/120 for zero action. Yet the policy remained 677–775 points below the survivor-conditioned statue mean.

This does not prove the checkpoint is behaviorally ideal. It directly shows that “exceed the survivor-conditioned statue” is not a necessary test of improvement over zero action.

The proposed T-Rex gate would be:

```text
3244.039 × 1.055 = 3422.461
```

The hashed run's best recorded 30-episode evaluation was `2549.69 ± 51.96`. The proposed gate would reject every checkpoint in the 6M-step curve by a large margin.

Runtime for the direct validation: clean `48fd90a`, macOS arm64, Python 3.12.13, MuJoCo 3.10.0, Gymnasium 1.3.0, NumPy 2.4.6, SB3 2.9.0, Torch 2.13.0, deterministic CPU evaluation. No training was run.

### Why `standing_mean + k × standing_std` is not the fix

It behaves better than `× 1.055` under zero, negative, and additively shifted rewards, but:

1. it still uses the wrong survivor-conditioned estimand;
2. episode SD measures dispersion, not uncertainty in the mean;
3. it does not shrink as sample size grows (`n=1` gives population SD 0);
4. it ignores the probability of entering the failure mode;
5. a universal `k` has no demonstrated meaning across species.

If a standing statistic is retained, it should be a labeled diagnostic. If “beat a surviving statue” is deliberately required, compare the policy's full-horizon-conditioned reward to that target and test survival separately.

### Recommended reward decision

For the scalar required by current consumers:

```text
G_stat = UCB95(mean unconditional zero-action return) + Δ_abs
G_run  = max(configured_literal, G_stat)
```

Then make the statistically fair confirmation:

```text
D_i = R_policy(seed_i) - R_zero(seed_i)
pass_paired = LCB95(mean(D_i)) >= Δ_abs
```

The full gate remains a conjunction:

```text
mean(R_policy) >= G_run
and pass_paired
and min_avg_episode_length
and every other enabled criterion
and required_consecutive
```

`Δ_abs` is a predeclared practical improvement in reward units. Use zero until there is evidence for a nonzero value; do not disguise an arbitrary effect size as 5.5%.

A scalar alone cannot preserve pairing and candidate uncertainty. `G_run` keeps downstream compatibility; the paired evidence makes advancement defensible.

### Required lifecycle

1. Materialize the fully effective reward/environment/perturbation/backend config.
2. Validate or measure a compatible baseline on a registered seed vector.
3. Resolve once and atomically persist `gate_resolution.json`.
4. Put the finite resolved float into an immutable run config.
5. Pass that same snapshot to SB3, JAX/MJX, notebook, reporting, visualization, result bundles, and sweeps.
6. Never let an executable consumer reopen raw TOML after resolution.
7. Resume loads the frozen gate and evidence state; it never recomputes in place.
8. A changed commit/config/backend is a new/forked run and recalibrates.

Missing, stale, corrupt, incompatible, non-finite, or underpowered baseline data must **block full curriculum advancement**. The literal may be used only through an explicit, persisted override or a diagnostic/pilot mode where advancement and publication-pass claims are disabled. It must not be a silent fallback.

Required provenance includes commit/ref/dirty patch, plant and policy-interface identity, reward and effective environment fingerprints, perturbation settings, backend, dependency versions, exact ordered seeds, reset protocol, per-episode outcomes, estimator/CI, command, capture time, and artifact hash.

### Perturbation handling

Separate two experiments:

- **perturbation off:** nominal plant-integrity control;
- **perturbation on exactly as advancement evaluation:** task-gate calibration.

Under nominal conditions, no survival may mean `CHECK PLANT` if passive stability is part of that species' plant contract. Under NS-1's intentional shove, a failing statue is expected, so `reward_mean_standing=None` must not trigger a plant diagnosis or literal fallback. Use unconditional, task-matched paired returns.

The current verdict ordering also masks Brachiosaurus's intended `CHECK PLANT`: it tests `gate <= reward_mean` before `n_standing == 0`.

### Current integration findings

- SB3 automated curriculum uses reward + length + sample count + `required_consecutive`.
- The notebook and main JAX artifact path perform one-shot checks.
- Legacy `jax_curriculum.py` checks reward only and passes when the gate is absent.
- Sweep and result-bundle decisions also have weaker/different semantics.
- Baseline and policy “publication seed” paths are not currently paired: baseline reseeds every episode; `eval_policy()` seeds once and advances an RNG stream.
- `collapse_peak_floor` still generically falls back to `min_avg_reward`, but current T-Rex stage 1 already sets `2200.0`. The comment's “no config sets it” claim is contradicted for current T-Rex; the other eleven stage configs remain coupled.
- Single-stage pilots still install plateau/collapse callbacks, so “advancement disabled” does not make a provisional reward threshold inert where collapse inherits it.

Name advancement, publication, sweep, and collapse decisions separately even when they share a scalar.

### PR #471 recommendation

1. Rebase to current `main` and remove both Python files; merged #472 supersedes them and current merge status is `CONFLICTING`.
2. Replace Part 2 §9's standing-floor formula and silent fallback with the lifecycle above.
3. Make `reward_mean_standing` diagnostic only.
4. Separate perturbation-off plant sanity from perturbation-on task calibration.
5. Add an explicit advancement-disabled pilot mode.
6. Make collapse configuration independent and update the current-T-Rex claim.
7. Correct “all four gates” to “all four reward sub-thresholds”; on the cited block only Velociraptor clears the complete reward+length gate.
8. Correct OQ-6's current Brachiosaurus reward-impact claim: the named weights are absent/default to 0 on the current stage-1 path.
9. Attach full provenance and raw artifacts to every retained NS-1/CEM/CoP measurement, or mark it unverified; reward-valued measurements are stale after later reward changes.
10. Implement the resolver in a separate tested PR/ADR. Green #471 CI cannot validate a resolver that is not in this diff.

If the full resolver is not ready, the smallest safe improvement is a **blocking zero-action preflight**: reject invalid/incompatible evidence, run the entire joint predicate against zero action, block a full curriculum when zero action can pass it, and allow bypass only in an explicit provenance-recorded diagnostic mode.

**Recommendation:** do not merge #471 as-is. Either rebase it to a substantially corrected documentation-only investigation, or close it as superseded and move the corrected gate design into a clean ADR plus implementation PR.

kuds:
## Response to the adversarial follow-up: three of my claims were wrong, and the proposed gate design supersedes mine

I verified the criticisms against the repository rather than accepting them. **Three hold and one of them invalidates the central recommendation in my consolidated §9.** I have one substantive disagreement, which is about scope rather than correctness.

---

## Conceded

### 1. The estimand mismatch — my §9 formula is wrong

Gating the policy's **unconditional** mean episode return against the statue's **survivor-conditioned** mean compares two different quantities. `curriculum.py:240` tests `latest["mean_reward"]`, which includes the policy's failures; `reward_mean_standing` averages only the statue's successes. Conditioning removes precisely the failure mode a trained controller is supposed to eliminate.

The paired experiment settles it — 120 seed-matched episodes, policy beating zero action by **+568.02** with survival **118/120 against 68/120**, while sitting 677–775 below the conditioned statue. `standing × 1.055` would reject a policy that is unambiguously better than doing nothing. That is a conceptual error, not a calibration disagreement. **Disregard §9's formula; use `UCB95(unconditional zero-action mean) + Δ_abs` with the paired LCB confirmation.**

Worth recording that the three seed blocks *confirm* the sampling-noise argument that motivated my formula, while showing the fix lies elsewhere:

| | block 3042 | block 4042 | block 5042 | spread |
|---|---:|---:|---:|---:|
| unconditional mean | 1971.57 | 1968.72 | 1884.71 | **86.9** |
| standing mean | 3244.04 | 3250.45 | 3233.99 | 16.5 |

The raw mean is indeed too noisy to calibrate against directly. I concluded the answer was to switch estimand; the correct answer is to keep the estimand and put a confidence bound on it. `UCB95` does that without importing the conditioning bias.

### 2. `collapse_peak_floor` — verified, my claim was false

```
configs/trex/stage1_balance.toml:164:collapse_peak_floor = 2200.0
```

I carried "no config sets it" over from the block deleted in `41f950b` without re-checking against current config. The coupling argument holds for the other eleven stage configs; the trex-specific claim did not.

### 3. "All four gates are cleared by a statue" — verified, and overstated

The gate is a conjunction and I tested only the reward half. Re-running with both criteria:

```
species              mean     len   gate  len gate   reward?   length?   FULL GATE
trex               1971.6   638.1   1840       750    CLEARS    blocks   blocked
velociraptor       1704.9   977.5    100       750    CLEARS    CLEARS   STATUE PASSES
brachiosaurus       108.2    98.7    100       750    CLEARS    blocks   blocked
dibothrosuchus     1702.0   674.2    100       750    CLEARS    blocks   blocked
```

Correct statement: **every stage-1 reward sub-threshold is cleared by a statue; only velociraptor's complete gate is.** Part 5's headline claimed more than the data supports.

The finding reframes rather than dissolves. `min_avg_episode_length` is doing all of the gating work in three of four species, the reward threshold is decorative in every one of them, and velociraptor's stage 1 can be passed outright by a do-nothing policy. It also means the notebook pre-flight is itself incomplete — it evaluates only `_gate <= reward_mean` and never the conjunction, so its verdicts describe one sub-threshold rather than advancement. The recommendation to run the entire joint predicate against zero action is correct and I had not proposed it.

### 4. Brachiosaurus OQ-6 weights — verified

`foot_contact_gate` and `foot_contact_weight` are absent from `configs/brachiosaurus/stage1_balance.toml` and default to `None`, so those sensors do not feed a reward term on the current stage-1 path. The sensor defect is real; the reward-impact claim attached to it is not, and I repeated it.

---

## One disagreement: the proposed gate does not test what this PR is about

The design correctly measures **improvement over the null**, and it is right that my formula conflated that with a different question. But the different question is the one #471 exists to answer, and the conjunction does not reach it.

Take the checkpoint from this very run — the one Part 4 shows is bouncing rather than standing (`bilateral_support_duty` 0.730, `single_support_duty` 0.061, `unsupported_duty` 0.209), and which scores 747.82 below a statue on the standing subset. Against the proposed gate:

| criterion | value | result |
|---|---|---|
| `UCB95(zero-action mean) + Δ_abs` (Δ=0) | ≈ 2359 vs policy 2496.22 | **passes** |
| paired LCB95 of improvement | +114.69 to +269.74 | **passes** |
| `min_avg_episode_length` | 1000.0 vs 750 | **passes** |
| `required_consecutive` | satisfied from ~3.5M | **passes** |

The bouncing policy clears the entire conjunction. That is not a hypothetical — it is the artifact this PR was opened over.

This is not an argument for my formula, which was wrong for the reasons above. It is an argument that **the statistical fix and the F1 fix are separate, and the comment's framing could be read as the former subsuming the latter.** "Beat a null policy on paired seeds" and "demonstrate balance rather than survival" are different predicates. NS-1's task change is what supplies the second, which is the thesis of this PR and remains untouched by either gate design. I would state that explicitly in any ADR so nobody ships the resolver believing stage 1 is then measuring balance.

## Minor

- The objection that episode SD measures dispersion rather than uncertainty in the mean is technically right but mischaracterises the proposal — I selected dispersion deliberately as an effect-size statement ("beats essentially every individual statue episode"), not as an uncertainty estimate. Objections 1, 3, 4 and 5 stand regardless, and 4 (ignoring the probability of entering the failure mode) is the strongest of them.
- `G_run = max(configured_literal, G_stat)` reinstates the stale literal as a permanent binding floor. A stale-**high** literal would then block a valid policy — the same failure this thread is about, inverted. I would prefer the literal be reachable only when no compatible baseline exists, recorded as such, rather than a standing ratchet.
- I could not verify the claim that `eval_policy()` seeds once and advances an RNG stream while the baseline reseeds per episode. It is plausible and it matters for the paired test; it should be confirmed before the resolver depends on it.

## Agreement on disposition

No disagreement on the PR itself: rebase to documentation-only, drop `zero_action_baseline.py` and `test_zero_action_baseline.py` as superseded by `3a8a37a`, correct the NS-1 figures for the post-`435f35f` floor, and move the resolver into its own tested PR or ADR rather than validating it against this diff's CI.

My consolidated comment should be read with §9's formula struck and §3's headline narrowed to the reward sub-threshold. The lifecycle half of §9 — resolve once, freeze, persist provenance, never reopen raw TOML, block rather than silently fall back — I still consider correct, and it is strengthened by the version proposed above.

---

kuds:
## Closing — the code half already landed, and the branch can no longer be merged

Closing this in favour of porting its documentation forward onto current `main`. The design work here was right and is preserved; the branch itself is the problem.

### Why it can't be merged

This branch is based on `ec2312543718`, before the structural refactors in #470–#477. Measured against current `main` (`6b1ad6e`):

```
125 files changed, 11234 insertions(+), 15459 deletions(-)
```

Merging it would **revert eight merged PRs** — un-splitting `curriculum/`, `plant_contract/`, `reporting/` and `result_bundle/` back into single modules, deleting `stance_diagnostics.py`, and reverting `trex_env.py`, `evaluation.py`, `jax_eval.py` and the CI workflow.

### The code contribution is already on `main`

The `zero_action_baseline` NaN fix landed independently in `3a8a37a`:

- `environments/shared/scripts/zero_action_baseline.py:125` returns `None` rather than `float("nan")`
- `environments/shared/tests/test_zero_action_baseline.py` is present, with the strict round-trip that rejects bare `NaN`/`Infinity`

Verified with `git merge-base --is-ancestor 3a8a37a origin/main`. Re-merging would only reintroduce the pre-refactor version.

### What was still outstanding, and where it went

Only the `TREX_REVIEW_2026_07.md` rewrite — NS-1 restated as *"change the task, not the reward"*, plus OQ-6. That is now on `claude/repo-review-stage-1-split-9jio88`, ported with four corrections established after this branch was written:

| item | correction |
|---|---|
| Correction 2 — "ship fixed, not ramped, because `set_reward_weight` is a bare `setattr`" | **Withdrawn.** `RewardRampCallback` already interpolates from global timesteps; the setter applies what it computed. The real gap is a dynamic perturbation-scale input with one unit across backends. |
| Correction 4 — "force the perturbation off in every diagnostic script" | **Superseded** by explicit `plant_sanity` / `task_gate` modes; forcing it off recreates the incompatible-baseline problem. Also listed `actuator_saturation_report` in error — it loads raw XML and never builds an env from TOML. |
| OQ-6 reward-path claim | **Does not hold on current `main`.** `configs/brachiosaurus/stage1_balance.toml` sets neither `foot_contact_gate` nor `foot_contact_weight`, so no stage-1 reward term reads those sensors on either affected species. The live impact is on the **observation**. |
| OQ-6's unverified velociraptor number | **Measured, and it was right: 0.553.** Cause identified — the touch site sits on `toe_d3` alone, missing `metatarsus` (17.54 N) and `toe_d4` (12.03 N) per foot. |

The pushed-task table also now carries an explicit staleness warning: those figures cannot be reproduced from this repository, predate `435f35f`, and their two binomial readings are corrected (`0 of 40` bounds survival **above** by 7.216%; the checkpoint's 85% is 34/40 with a 0.72526 lower bound, taken at noise 0.05 while the config retains 0.10).

### OQ-6 turned out to be broader than filed

Auditing all four species against `mj_contactForce` found a second affected species:

| species | sensor / measured contact | contact / weight |
|---|---|---|
| trex | 1.000 | 1.002 |
| dibothrosuchus | 1.000 | 1.000 |
| velociraptor | **0.553** | 1.000 |
| brachiosaurus | **0.000** | 0.988 |

`contact / weight` is 1.00 everywhere, so the plants are genuinely in equilibrium and these are sensor-scope defects rather than physics ones. Recorded in `docs/investigations/FOOT_SENSOR_VERIFICATION.md`, reproducible via `environments/shared/scripts/foot_sensor_report.py`.

Nothing here is lost — please reopen if I've misread what was still needed from this branch.

---

PR #474, opened 2026-07-30
Design proposal: split Stage 1 balance into stance (1a) and recovery (1b)

kuds:
## Summary

Design proposal to split the single Stage 1 "balance" stage into two independently gated stages:

- **Stage 1a — Stance**: settle from a randomised initial state and hold the pose to the horizon.
- **Stage 1b — Recovery**: hold that stance against scheduled external disturbance, and demonstrate *recovery*, not merely survival.

Documentation only — one file, `docs/STAGE1_SPLIT_PLAN.md`. No code changes.

## Motivation

Stage 1 currently asks one reward threshold to answer two questions: whether the plant reached a stable stance, and whether that stance is actively controlled. The first is legitimately satisfied by a controller that settles and then stops working. The second is not measurable without a disturbance. On the undisturbed task — where a passive and an active controller can generate the same trajectory — realized on-trajectory return cannot identify active feedback. That is why the gate has been either unbindable or unclearable in every configuration tried.

Re-measured on `34a7002`: a zero-action policy still clears the **reward** threshold on all four species, and still clears velociraptor's **complete** reward-plus-length gate outright (1704.93 mean against a gate of 100; 977.5 mean length against 750).

## Revision 4 — rebased onto the tree where the prerequisites landed

Revision 3 identified five blocking defects and said they "must be fixed before any gate in this document can mean anything." **They have been**, in #478:

| § | defect | status on `34a7002` |
|---|---|---|
| 7.5 | ~1% of episodes terminal at generation | fixed — 43/4000 → **0/4000**, `71a91b7` |
| 5.2 | gate plumbing fails open on both backends | fixed — both reject, `17ca4e5` |
| 7.4 | `collapse_peak_floor` inherits `min_avg_reward` | fixed — decoupled, `17ca4e5` |
| 7.1 | `[0, 0]` scores airborne as perfectly balanced | fixed — reports 1.0, `c667938` |
| 7.2 | foot touch sensors unverified | done — audited, `2ab5ae6` / `efe93dd` |

§7.3 is **not** claimed as landed: it needs a perturbation to switch off, which does not exist yet.

Each section keeps its original diagnosis — that reasoning is what justifies the gate design — and gains a **STATUS** note recording what shipped and where it differed from the proposal.

## The two substantive consequences

**§2.3.1's central empirical claim is stale.** It rejected `LCB95 ≥ 0.90` at n=40 partly because the current checkpoint scored 39/40 on two of four panels. But §7.5 identified that first failure — **seed 3077** — as the reset defect itself. Verified on `34a7002`: seed 3077 now spawns at 0.72000 m and survives. The panels were contaminated by an environment bug, the checkpoint may be 40/40 throughout, and the case for `n = 179` may evaporate entirely. The power table is arithmetic and stands; the premise that a good policy fails the cutoff does not. §10 now sequences the re-baseline **first**, before any operating point is declared.

**§7.2 found two defects nobody had filed.** Velociraptor's foot sensors report **55.3%** of true load (the site sits on `toe_d3` alone, missing the metatarsus and lateral digit); brachiosaurus's report **0.000**. `contact/weight` is 1.00 on all four species, so these are sensor-scope defects rather than physics ones. Neither reaches a stage-1 reward term today, but both reach the **observation** — brachiosaurus trains with four permanently dead input channels. This forces T-Rex-first enablement by instrumentation rather than preference, adds a risk row, and gives brachiosaurus's existing risk row a measured partial explanation that does not require plant corruption.

## Other corrections in revision 4

- **§6.2 read its own GRF invariant backwards.** The statue's 841 N is *exactly* body weight (840.9 N, ratio 1.002). The 1483 N that made it look low is `mj_getTotalmass` counting the **65.45 kg prey body** — 43% of the model total. Any implementation must divide by the animal's kinematic subtree or it will fire on every species carrying prey.
- **§1.5's sensor escape route is closed in both directions.** T-Rex's sensors agree with `mj_contactForce` to 1.000 statically, and the duty classifier tracks kinematic ground truth to within 0.52% of steps across a swept 3%–67% airborne range. `unsupported_duty = 0.209` can be taken at face value; only the *causal* claim remains inferred.
- **All 12 code citations remapped**, each verified to resolve to the exact line it describes. The four monoliths they pointed at were split into packages in #472–#477.
- **§11 claim ledger** now separates `48fd90a` from `34a7002` measurements and lists what became stale.

## Review history

Revision 4. Three rounds of review plus an empirical addendum; every checkable claim was verified against the repository before being accepted, and several corrected claims this document originally got wrong.

Both surrounding-documentation risks revision 3 flagged are now closed: `TREX_REVIEW_2026_07` §NS-2 is explicitly superseded on `main` (`3774301`), and #471 is closed with its NS-1 rewrite ported forward and four of its own corrections withdrawn or narrowed.

## What is still outstanding

The split itself. §2.3's episode-level gate metrics, §3's perturbation scheduler, §4's stage manifest, and §5's gate resolver have not started. §10 gives the order, with the re-baseline first.


kuds:
## Adversarial follow-up: approve the stance/recovery split, revise the gate contract before implementation

### Bottom line

I agree with the central premise of this PR. Splitting Stage 1 into:

- **stance acquisition under nominal conditions**, and
- **recovery under an explicit external disturbance**

is a materially better task design than trying to infer active balance from an undisturbed reward trace. An intervention is what makes disturbance rejection observable.

I would nevertheless revise this document before treating it as an implementation contract. The remaining issues are not merely code details: the proposed success estimands, statistical decision rules, task compatibility, perturbation calibration, and stage migration are still underspecified or internally inconsistent.

Snapshot reviewed: PR head `f5ed6a1a9eed607e42af6dcfa0b360cfb4a1f286`, current/base `main` `48fd90afc1bef16f7f45054f31571cd99dfc6fb1`. The diff is documentation-only and current CI is green.

### What is sound

- Dropping `reward_mean_standing × 1.055` is correct. It compared unconditional policy return with zero-action return conditioned on full-horizon survival.
- Stage 1a should not use return to distinguish active from passive standing.
- Registered, identical seed schedules and paired policy-versus-baseline comparisons are the right direction.
- Resolve once, freeze, persist, validate on resume, and fail closed on stale/missing calibration is the right lifecycle.
- The sensor check and `collapse_peak_floor` decoupling are real prerequisites.
- §8 correctly identifies the missing load-bearing measurement: the current checkpoint at noise `0.10` under the finalized pushed task.

### Findings

#### 1. `0/40` is strong evidence, not a zero null

[§3.4](https://github.com/kuds/mesozoic-labs/blob/f5ed6a1a9eed607e42af6dcfa0b360cfb4a1f286/docs/STAGE1_SPLIT_PLAN.md#L223-L246) says the survival threshold needs no confidence-interval argument because the pushed statue achieved 0/40 full horizons.

For 0 successes in 40 Bernoulli episodes, the exact one-sided 95% upper confidence bound is:

```text
1 - 0.05^(1/40) = 0.07216
```

The defensible statement is therefore: under that exact sampled task, the evidence bounds zero-action survival below about 7.2%, not at zero. That still provides strong practical separation from 70%.

Candidate uncertainty matters too. The cited checkpoint result is 85%, or 34/40; its exact one-sided 95% lower bound is about 72.5%, only narrowly above the proposed 70%. It was also measured at noise `0.05`, while the intended configuration retains `0.10`, where the checkpoint is unmeasured.

The `0.70`, `800`, and three-consecutive thresholds should be marked provisional until the final pushed task is measured. Repeating the same deterministic seed panel three times establishes process stability, not three independent statistical confirmations.

#### 2. The calibration instructions contradict each other

Stage 1b says to resolve against the pushed floor, and §5 correctly requires the fully effective perturbation configuration. But [§7.3](https://github.com/kuds/mesozoic-labs/blob/f5ed6a1a9eed607e42af6dcfa0b360cfb4a1f286/docs/STAGE1_SPLIT_PLAN.md#L373-L377) tells `zero_action_baseline` to force `perturbation_delta_v = 0.0`.

That recreates the incompatible-baseline problem. Diagnostic/evaluation tooling needs explicit task modes:

```text
plant_sanity : perturbation forced off
task_gate    : perturbation exactly matches full-strength advancement evaluation
```

The mode and complete perturbation fingerprint must be persisted. A tool must never silently turn the configured task off.

The same paragraph also repeats a false claim from #471: `actuator_saturation_report.py` does not build an environment from TOML; it loads raw XML and steps MuJoCo directly ([current code](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/scripts/actuator_saturation_report.py#L44-L76)).

#### 3. The Stage 1a gate is not defined precisely enough to implement

[§2.3](https://github.com/kuds/mesozoic-labs/blob/f5ed6a1a9eed607e42af6dcfa0b360cfb4a1f286/docs/STAGE1_SPLIT_PLAN.md#L153-L168) proposes `min_full_horizon_share`, `max_avg_height_error`, and `max_avg_drift_distance`. Current advancement supports only reward, mean length, velocity, success rate, evaluation count, and consecutive passes ([`StageThreshold`](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/curriculum.py#L75-L84), [decision logic](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/curriculum.py#L239-L260)).

Logging related values does not make them canonical gate evidence. Define:

- per-step to per-episode to evaluation aggregation;
- whether failed episodes contribute equally;
- whether drift means final, maximum, time-average, or cumulative displacement;
- the settling/tail window;
- species-specific or morphology-normalized thresholds;
- SB3 and MJX/JAX parity;
- report, bundle, notebook, and sweep persistence.

`height_error` is currently T-Rex-specific instrumentation, not a four-species advancement metric. Average height and average drift can also hide oscillation and large transient failures.

There is a naming ambiguity as well: literal zero action does not pass the proposed T-Rex Stage 1a thresholds—it measured 57.5% full-horizon against the proposed 90%. What may legitimately pass is a controller that corrects reset randomization and then becomes passive. That is not the same null.

#### 4. Load compatibility is not task compatibility

[§3.2](https://github.com/kuds/mesozoic-labs/blob/f5ed6a1a9eed607e42af6dcfa0b360cfb4a1f286/docs/STAGE1_SPLIT_PLAN.md#L185-L210) concludes that a `step()` hook “invalidates no existing checkpoints” because `step` is not included in `policy_interface_revision`.

The narrow fact is verified: the current contract fingerprints observation/action implementations and selected reset semantics, not `BaseDinoEnv.step` ([current contract](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/plant_contract.py#L895-L921)).

The conclusion is too broad. A pushed task changes the transition kernel and evaluation distribution. Existing checkpoints may remain mechanically loadable and policy-interface compatible while being unvalidated for the new task. The absence of `step` from the fingerprint is a provenance gap, not evidence of task equivalence.

Add a distinct task/evaluation fingerprint covering perturbation implementation, schedule, RNG protocol, force parameters, reset configuration, horizon, reward/termination semantics, and backend. Resume must not cross that boundary silently.

#### 5. The replacement reward rule mixes two decision procedures and revives the stale literal

[§5](https://github.com/kuds/mesozoic-labs/blob/f5ed6a1a9eed607e42af6dcfa0b360cfb4a1f286/docs/STAGE1_SPLIT_PLAN.md#L288-L313) proposes:

```text
G_stat = UCB95(mean unconditional zero-action return) + Δ_abs
G_run  = max(configured_literal, G_stat)

D_i         = R_policy(seed_i) - R_zero(seed_i)
pass_paired = LCB95(mean(D_i)) >= Δ_abs
```

The paired comparison is the statistically coherent superiority test. If the unpaired scalar and paired test are both mandatory, the scalar can reject a candidate whose paired improvement is precise and positive, discarding the covariance benefit that motivated pairing.

`max(configured_literal, ...)` also carries a legacy, task-dependent number into a new pushed task. A literal may be an explicit, fingerprinted safety requirement; it must not be an implicit fallback or unexplained operand.

Recommended distinction:

```text
D_i = R_policy(seed_i) - R_zero(seed_i)
pass_reward = LCB95(mean(D_i)) >= Δ_R       # authoritative, if reward is retained

G_screen = UCB95(E[R_zero]) + Δ_R           # optional scalar for display/screening
```

`G_screen` should not independently override the paired decision. With `Δ_R = 0`, this establishes statistical superiority only, not practical usefulness; a nonzero effect size needs a task-based rationale.

For Stage 1b, reward should be secondary to directly measured recovery capability anyway.

#### 6. The perturbation scheduler and ramp are not yet a backend-neutral design

A shared `external_push_force()` helper can share force arithmetic, but not scheduler state or simulator mutation. The design must specify:

- explicit clearing of `xfrc_applied` after a pulse and on reset;
- deterministic episode-local push times and directions;
- pre-generated schedules so baseline and policy receive identical disturbances;
- MJX `data.replace(xfrc_applied=...)` and auto-reset clearing;
- CPU JAX evaluation parity;
- a common schedule unit across SB3 and thousands of parallel MJX environments;
- persisted/restored ramp progress on resume;
- calibration and advancement always at frozen full strength, never partway through the training ramp.

[§3.3](https://github.com/kuds/mesozoic-labs/blob/f5ed6a1a9eed607e42af6dcfa0b360cfb4a1f286/docs/STAGE1_SPLIT_PLAN.md#L212-L221) also attributes the ramp problem to `set_reward_weight` being a bare setter. The current `RewardRampCallback` already calculates interpolated values from global timesteps and periodically propagates them; the setter itself does not force a step function ([current code](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/curriculum.py#L794-L877)). The missing piece is a dynamic perturbation-scale input with one defined unit and resume behavior.

Ramp-versus-fixed remains a hypothesis. It should be selected by a small transfer pilot, not assumed to prevent catastrophic forgetting.

#### 7. Renumbering is a schema migration, not a mechanical edit

[§4](https://github.com/kuds/mesozoic-labs/blob/f5ed6a1a9eed607e42af6dcfa0b360cfb4a1f286/docs/STAGE1_SPLIT_PLAN.md#L252-L269) understates the cost. Current configuration loading, curriculum loops, JAX, reporting, result bundles, notebooks, sweeps, checkpoint paths, generated website data, and historical summaries encode exactly stages 1–3. For example, config discovery hard-codes three prefixes ([`config.py`](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/config.py#L151-L169)) and bundle reporting accepts only stages 1–3 ([`reporting.py`](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/reporting.py#L962-L974)).

Renumbering silently changes the historical meaning of “stage 2” and “stage 3.” Prefer stable semantic IDs:

```text
stance
recovery
locomotion
behavior
```

with a separate display/order field, a schema-version bump, and backward readers for existing three-stage artifacts. Start with T-Rex-only experimental enablement until each species has task-matched evidence.

#### 8. Several empirical statements still need narrower labels and stronger provenance

- “The policy is hopping” and “the reward caused it” are hypotheses until the touch sensors agree with `mj_contactForce` and kinematic flight phases. The `[0,0]` arithmetic hole is verified; behavioral causation is not.
- “Brachiosaurus cannot stand at all” overstates a 0/40 zero-action result under randomized reset. It does not show that no learned controller can stand, nor by itself prove plant corruption.
- The grouped provenance section does not attach each load-bearing number to a command, exact effective config, backend, dependency versions, ordered seeds, raw episode data, and artifact hashes.
- The pushed reward values are explicitly stale, while the survival counts remain unverified because the producing perturbation implementation and raw schedule are absent from the repository.

Please add a claim ledger or immutable measurement manifest and label each retained claim `measured`, `artifact-derived`, `inferred`, `stale`, or `unverified`.

### Recommended gate contract

For **Stage 1a**, perturbations are off and reset randomization is fingerprinted. Define an episode-level event:

```text
stance_success =
    full_horizon
    and settles_by <= T_settle
    and tail_q95(height_error) <= h_max
    and tail_q95(orientation_error) <= angle_max
    and tail_q95(planar_speed) <= v_max
    and tail_q95(angular_speed) <= omega_max
    and tail_drift_rate <= d_max
```

Require a one-sided lower confidence bound on the stance-success probability. A controller may become passive after settling; literal zero action need not pass.

For **Stage 1b**, training may ramp only if a pilot supports it, but every calibration and advancement evaluation uses the same finalized full-strength perturbation distribution. Define per-shove recovery as re-entering a safe pose/velocity/contact set within `T_recover` and remaining there for a dwell interval. Require:

```text
LCB95(P(full horizon and every shove recovered)) >= p_recovery
LCB95(mean(policy_success_i - zero_success_i)) >= Δ_success
```

Optionally retain paired unconditional reward as a secondary criterion:

```text
LCB95(mean(R_policy_i - R_zero_i)) >= Δ_R
```

The null suite should include zero action, constant/brace policies, and the incoming Stage 1a checkpoint. Survival alone does not prove feedback control.

### Smallest useful sequence

1. Verify touch sensors against `mj_contactForce`.
2. Decouple collapse detection and set explicit per-stage behavior before removing reward gates.
3. Define task fingerprints and semantic stage IDs/backward compatibility.
4. Implement and parity-test the new episode-level gate metrics.
5. Implement a deterministic perturbation scheduler with force-off regression, force clearing, seed/schedule, SB3/MJX, and resume tests.
6. On T-Rex, evaluate zero action, constant/brace controls, and the existing checkpoint at noise `0.10` under the finalized full push on the same registered 40-seed schedule, then repeat on a held-out block.
7. Calibrate thresholds and decide fixed versus ramped training.
8. Enable recovery per species only after its own plant and learnability preflights.
9. Implement the resolver before, or atomically with, any executable stage that depends on it.

No new training is needed for the prerequisite transfer tests. The current push evaluation cannot be reproduced from this PR because the perturbation implementation, exact schedule, and raw outcomes are not present.

### Recommendation

Keep this PR as the design proposal—it captures the right architectural correction—but revise the claims and contract above before merge. My disposition is:

> **Conceptual design approved; gate/statistical/task-compatibility specification needs changes before implementation.**

kuds:
## Revised in `9e9924f` — all findings accepted, five code citations and two bounds independently verified

I verified every checkable claim against the repository before revising rather than taking them on trust. All five code citations and both statistical bounds hold. Seven of the findings identify things this document got **wrong**, not merely underspecified, and three of those were errors it inherited from §NS-1 or from my own PR #471 comments.

### Verified before accepting

| claim | check | result |
|---|---|---|
| `StageThreshold` supports only 6 fields | `curriculum.py:79-84` | confirmed — no full-horizon/height/drift fields exist |
| `RewardRampCallback` already interpolates | `curriculum.py:794-877` | confirmed — "linearly increases it over `ramp_timesteps`" |
| `actuator_saturation_report` uses raw XML | `scripts/actuator_saturation_report.py:44-76` | confirmed — `mujoco.MjModel.from_xml_string`, no TOML env |
| `config.py` hard-codes three stage prefixes | `config.py:154` | confirmed — `{1:…, 2:…, 3:…}`; stage 4 raises `KeyError` |
| bundles reject stages outside `{1,2,3}` | `reporting.py:968` | confirmed |
| 0/40 one-sided 95% upper bound | exact binomial | **0.07216** |
| 34/40 one-sided 95% lower bound | exact binomial | **0.72526** vs the proposed 0.70 gate |

### What changed

**§TL;DR, §2.1 — the null was mislabelled.** "A statue passes 1a by design" was imprecise in a way that mattered. Literal zero action reaches the full horizon in 57.5% of episodes, well under the proposed 90%. What passes is a *settle-then-passive* controller. Both stages now state this distinction explicitly.

**§3.4 — 0/40 is a bound, not a zero.** Replaced "needs no confidence-interval argument" with the exact bound (≤ 7.216%). Recorded that the 85% candidate figure is 34/40 with a 0.72526 lower bound — barely above a 0.70 gate — and was taken at noise 0.05 while §2.2 retains 0.10. Added that repeating a deterministic seed panel three times is process stability, not three independent confirmations. All 1b thresholds now marked provisional.

**§3.3 — the ramp rationale was wrong.** `RewardRampCallback` already interpolates from global timesteps; the setter is not the blocker. §NS-1 correction 2 is wrong on the mechanism and revision 1 repeated it. The real gap is a dynamic perturbation-scale input with one unit across backends and defined resume behaviour. Ramp-versus-fixed is now an open question for the §8.1 pilot rather than an assumption.

**§7.3 — the calibration contradiction.** Forcing `perturbation_delta_v = 0.0` in diagnostic tooling recreated the incompatible-baseline problem it was meant to prevent, since 1b's gate must be calibrated against the *pushed* floor. Replaced with explicit `plant_sanity` / `task_gate` modes, both fingerprinted and persisted. Removed `actuator_saturation_report` from the affected list.

**§5.1 — the gate rule mixed two decision procedures.** You're right that mandating both an unpaired scalar and a paired test lets the scalar reject a candidate whose paired improvement is precise and positive, discarding the variance reduction that motivates pairing. The paired test is now authoritative and `G_screen` is display/screening only. `max(configured_literal, …)` is gone — I had raised exactly that objection on #471 and then transcribed the formula into this document anyway.

**§3.2 — load compatibility narrowed.** The fingerprint fact is verified, but "invalidates no existing checkpoints" overstated it. Checkpoints stay mechanically loadable and interface-compatible while being unvalidated for a changed transition kernel. Added a task/evaluation fingerprint requirement and full scheduler requirements.

**§2.3 — 1a gate respecified.** Now an episode-level `stance_success` event using tail quantiles rather than means, gated on a lower confidence bound of its probability. Revision 1 said the quantities "are already logged," conflating diagnostic logging with canonical gate evidence; the aggregation semantics, tail window, drift definition, per-species normalisation, backend parity and persistence are now listed as work.

**§4 — semantic IDs replace renumbering,** with a schema bump and backward readers, plus T-Rex-only enablement first.

**§1.5, §9 — claims narrowed.** Hopping and its reward causation relabelled `[inferred]` pending sensor verification; the `[0, 0]` hole stays `[measured]`. The brachiosaurus risk row now says what the zero-action result does and does not show.

**§11 — claim ledger added,** labelling every load-bearing number `measured` / `artifact-derived` / `inferred` / `stale`, and stating plainly that the push-on figures cannot be reproduced from this repository until the perturbation implementation lands. Per-measurement manifests are noted as the next step rather than claimed as done.

**§10 — resequenced** along the lines you proposed: sensor check first, collapse decoupling before any reward gate is removed, then fingerprints and IDs, then metrics, then the scheduler, then measurement, then calibration.

### No pushback

I don't disagree with anything in the review. The one thing I'd underline for anyone reading later: the split does not by itself make Stage 1a measure balance — that remains F1's territory and the disturbance is what addresses it. §2.1 now says so directly rather than leaving it implied.

CI green on `9e9924f`.

---

kuds:
## Follow-up on revision 2 (`9e9924f`)

I re-reviewed the revised document against the previous head (`f5ed6a1`) and current `main` (`48fd90a`), including the response above. Revision 2 materially addresses most of the original review: the stance/recovery estimands are much better defined, the binomial bounds are corrected, the paired gate is authoritative, calibration modes are separated, checkpoint claims are narrowed, semantic IDs replace blind renumbering, and ramping is now treated as an open experimental question.

My updated disposition is:

> **Directionally sound and no longer in need of a redesign, but one more focused documentation pass is needed before this is an implementation-grade contract.**

I do not think all supporting code must land in this documentation-only PR. The document should, however, define the following contracts—or clearly mark them as blocking prerequisites—so that an older consumer cannot silently interpret the new configuration permissively.

### Remaining high-priority issues

1. **Make the new gate schema explicitly fail closed.**

   Removing `min_avg_reward` before every consumer understands the composite gate can auto-advance rather than fail. Current SB3 defaults omitted thresholds to permissive values, and current JAX returns `True` when `min_avg_reward` is absent ([SB3](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/curriculum.py#L79-L84), [JAX](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/jax_curriculum.py#L34-L50)). Require a versioned `gate_schema_version` / `gate_kind`, reject unknown gate types and fields, and land the config plus all executable consumers atomically.

2. **Finish the statistical decision contract.**

   The document still does not specify the interval method, advancement sample size, paired binary risk-difference estimator, calibration versus held-out panels, or multiplicity rule for the null suite. Under exact one-sided 95% Clopper-Pearson bounds:

   ```text
   29/30 -> LCB 0.85140
   30/30 -> LCB 0.90497
   39/40 -> LCB 0.88681
   40/40 -> LCB 0.92784
   ```

   Thus `LCB95(P(stance_success)) >= 0.90` requires a perfect panel at both common sample sizes. That may be intentional, but it must be stated. Also align the Stage 1b formula with the prose: either beat every predeclared null using simultaneous lower bounds, or select the strongest null on calibration seeds and confirm once on held-out seeds. The current equation compares only with zero action even though the suite also names brace/constant controllers and the incoming Stage 1a checkpoint ([§3.4](https://github.com/kuds/mesozoic-labs/blob/9e9924fc586ee72523601ec9ec34175006bd5b09/docs/STAGE1_SPLIT_PLAN.md#L307-L323)).

3. **Separate normative capability requirements from baseline-derived evidence.**

   Safe height, tilt, speed, drift, settling time, required recovery probability, and maximum recovery time are task requirements; they should not track whatever the null controller happens to do. The baseline should support the blocking preflight and relative-superiority comparisons. Replace “both stages resolve their thresholds from a measured baseline” with a lifecycle that freezes:

   - a versioned capability specification;
   - a compatible baseline/null manifest;
   - the predeclared statistical procedure and seed panels.

   See [§5](https://github.com/kuds/mesozoic-labs/blob/9e9924fc586ee72523601ec9ec34175006bd5b09/docs/STAGE1_SPLIT_PLAN.md#L353-L368).

4. **Distinguish resume from next-stage initialization.**

   “Do not cross a task fingerprint” conflicts with intentionally initializing Stage 1b from a Stage 1a checkpoint. Define:

   - `resume_same_stage`: exact task, resolved gate, scheduler/ramp state, optimizer and normalization compatibility required;
   - `initialize_next_stage`: task mismatch expected and recorded as lineage, policy-interface compatibility required, with explicit optimizer/normalization/ramp reset behavior.

   Exact mid-stage reproducibility also requires more than the current checkpoints preserve: PRNG, environment/scheduler state, registered schedule position, global transition count, and ramp state. Otherwise narrow the promise to reproducible stage-boundary restart.

5. **Turn semantic IDs and fingerprints into executable manifests.**

   Add a versioned per-species ordered manifest containing stable stage ID, config, terminal-capability flag, and legacy numeric alias; this is what makes T-Rex-only recovery representable without reinterpreting historical stages. Separate:

   - **task identity**: plant/policy-interface identity, model and implementation hashes, full effective environment/reward/termination/perturbation config, backend/precision and relevant dependencies;
   - **evaluation protocol**: null-controller definitions, ordered episode seeds, pair IDs, episode count, confidence procedure, and intended plus realized push schedule.

   Recovery evidence should emit push ID, actual start/end steps, vector/impulse, schedule hash, recovery-entry step, and dwell result.

6. **Update the stale surrounding documentation.**

   The PR description still describes revision 1: assumed ramping, numeric renumbering, broad checkpoint validity, proven hopping, and concrete calibrated thresholds. It should be updated to match `9e9924f`. The document also cites current-main `TREX_REVIEW` sections that still recommend the now-rejected survivor-conditioned standing floor ([current NS-1](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/docs/investigations/TREX_REVIEW_2026_07.md#L883-L907)). Update or explicitly supersede those sections so merging #474 alone does not leave contradictory canonical guidance.

### Specific corrections

- In the `0/40` paragraph, zero-action survival is bounded **above**, not “below,” approximately 7.216%.
- Removing both `collapse_peak_floor` and `min_avg_reward` produces a fallback of `0.0`, not an undefined value ([current code](https://github.com/kuds/mesozoic-labs/blob/48fd90afc1bef16f7f45054f31571cd99dfc6fb1/environments/shared/curriculum.py#L1002-L1017)). Decoupling first is still the correct recommendation; the failure mode is that collapse detection can arm after any positive robust peak.
- The resolver is listed after stage enablement even though the text says it must land before or atomically with the dependent stage. Reverse those steps.
- Rename the `FULL GATE` table: it reproduces the reward-plus-length conjunction on one 40-episode aggregate, not evaluation batching, minimum episode count, and `required_consecutive`.
- `perturbation_delta_v = 1.5` is dimensionally ambiguous if `1.5` is a capture-velocity multiplier. Name either the dimensionless multiplier or an actual `delta_v_mps`, and persist the derived force/impulse per species.
- Narrow the universal “no episode-return gate can separate” claim. The supported conclusion is that, on the undisturbed task, realized on-trajectory state reward cannot identify active feedback when passive and active controllers generate the same trajectory.

### Claim-response audit

The response correctly verified the two binomial bounds and the cited facts about `StageThreshold`, reward interpolation, raw-XML saturation reporting, and numeric stage assumptions. “All findings addressed” is nevertheless too strong: the universal reward-impossibility wording, complete statistical contract, complete provenance/lifecycle, and full-gate-table issue remain.

With those corrections, I would regard this as a defensible design proposal and would move the remaining task-matched measurements into explicit implementation prerequisites rather than requesting another architectural redesign.

### Targeted validation addendum

I ran bounded evaluation-only checks after the review. These sharpen two recommendations above; they do not validate the proposed pushed task.

**Snapshot and runtime.** PR head `9e9924f`; current `main` and evaluation code `48fd90a`; macOS 26.5.2 arm64; Python 3.12.13; MuJoCo 3.10.0; Gymnasium 1.3.0; Stable-Baselines3 2.9.0; NumPy 2.4.6. The committed T-Rex Stage-1 config SHA-256 was `9d36537e…fefb0` and the model XML was `c64fb09d…b628c`. No training was run and the repository remained clean.

#### 1. The current gate plumbing demonstrably fails open

Using a temporary composite-only TOML gate:

```toml
[curriculum]
gate_schema_version = 1
gate_kind = "stance_success_lcb"
min_stance_success_lcb = 0.90
```

the loader preserved the unknown fields, but `thresholds_from_configs` silently discarded them. SB3 materialized the legacy defaults (`min_avg_reward=-inf`, length/success floors zero) and returned `False, False, True` over three ten-episode evaluations whose reward was deliberately `-1e12` and length was `1`. The legacy JAX check returned `True`; the active JAX evaluation check returned `(True, [])`. Existing focused tests also passed because they currently codify permissive missing-threshold behavior.

This is measured confirmation that schema + SB3/JAX consumers + parity tests must land atomically. A missing, unknown, or incomplete gate must be fatal whenever advancement is enabled; absence is acceptable only in an explicit non-advancing diagnostic/pilot mode.

#### 2. Current-main zero-action reward numbers are stale while the physical result is reproducible

Exact command:

```bash
python environments/shared/scripts/zero_action_baseline.py \
  trex --episodes 40 --seed 3042
```

Two identical runs reproduced:

```text
reward mean          1971.57
reward standing      3244.04
full-horizon         23/40
mean length          638.1/1000
```

The historical physical result (`23/40`, mean length `638.1`) is unchanged, while the cited reward values (`1743.73` unconditional and `2834.95` standing) are no longer current after later reward changes.

Three disjoint additional panels starting at seeds `4042`, `5042`, and `6042` yielded `23/40`, `22/40`, and `25/40` survival. Pooled zero-action survival was `93/160 = 58.125%` (exact two-sided 95% Clopper-Pearson interval `50.08%–65.87%`); per-panel reward means ranged from `1884.71` to `2157.53`. This directly supports using state capability—not a historical reward scalar—as the primary stance gate.

#### 3. The proposed 40-episode LCB rule rejects the current checkpoint

I evaluated the available current-main artifact deterministically on the same four seed panels, paired with zero action:

```text
checkpoint SHA-256   ccb6f322…50549
VecNormalize SHA-256 3a636790…c9e1
training commit      48fd90a
training steps       5,950,000
reset noise          0.10
task                 unpushed Stage 1
```

Checkpoint survival was:

```text
seeds 3042–3081   39/40   CP one-sided LCB95 = 0.88681   FAIL proposed gate
seeds 4042–4081   39/40   CP one-sided LCB95 = 0.88681   FAIL proposed gate
seeds 5042–5081   40/40   CP one-sided LCB95 = 0.92784   PASS
seeds 6042–6081   40/40   CP one-sided LCB95 = 0.92784   PASS
pooled             158/160 CP one-sided LCB95 = 0.96118   PASS
```

Thus an observed 98.75%-reliable current checkpoint fails half of the proposed 40-episode panels because `LCB >= 0.90` permits no failures at `n=40`. This is not just a theoretical discreteness problem.

The first failure, seed `3077`, was already terminal immediately after `reset`, before any action: pelvis height `0.65386 m` was below the configured `0.70 m` termination floor. A deterministic scan of seeds `3042–7041` found `43/4000 = 1.075%` already-terminal resets (exact two-sided 95% interval `0.779%–1.445%`), all for the same height-floor reason.

Add a blocking reset-validity preflight. Reset generation must produce a nonterminal initial state, or use deterministic constraint-aware resampling whose attempts and realized state are recorded. Already-terminal task generation must not be counted as policy failure or silently discarded after outcomes are observed.

#### 4. Power the confirmation test; do not multiply confidence by `required_consecutive`

Exact one-sided 95% Clopper-Pearson cutoffs for `LCB >= 0.90` are:

```text
n=30   require 30/30
n=40   require 40/40
n=80   require at least 77/80
n=100  require at least 96/100
```

At true reliability `p=0.95`, one-panel pass probabilities are only `21.5%`, `12.9%`, `42.8%`, and `43.6%`, respectively. At `p=0.98`, they are `54.5%`, `44.6%`, `92.3%`, and `94.9%`.

If a 95%-reliable policy should advance with roughly 80% probability, the exact minimum is `n=179` with at least `168` successes; a round `n=180` uses cutoff `169/180`. If the intended capability is instead near-perfect `p=0.98`, `n=100`, cutoff `96/100`, is reasonable. The PR needs to declare that operating point.

Use cheap evaluations and `required_consecutive` only as scheduler hysteresis. Once qualified, freeze the candidate and run one predeclared held-out confirmation panel. Reusing the same deterministic panel three times adds no statistical evidence; requiring three independent panels to pass discards information and has very poor power.

#### 5. Pair identity is part of the estimand

A fixed-seed synthetic check demonstrated why marginal totals are insufficient. Two paired panels both had policy `30/40`, baseline `20/40`, and marginal improvement `0.25`; changing only which seeds succeeded changed an illustrative paired normal lower bound from `0.136` to `0.032`. In a 100,000-trial correlated-reward simulation, ignoring within-seed covariance made a nominal 5% boundary test reject anywhere from `0.001%` to `9.08%`, while the paired calculation stayed near `5.4%`.

Therefore every canonical gate record needs per-controller pair ID, episode seed, success outcome, return, and realized push schedule. An aggregate CSV containing only marginal means cannot reproduce the paired decision.

#### 6. Pushed-task claims remain unverified

The available checkpoint and `evaluations.npz` are mechanically evaluable but contain no perturbation events, episode seeds, recovery outcomes, or raw trajectories. The repository still lacks the proposed scheduler, authoritative force conversion, registered schedule, realized per-episode pushes, and raw pushed outcomes. The available checkpoint also differs from the older artifact cited for the published `2489.65` comparison.

Consequently, the reported pushed returns and `0/40` result remain artifact-derived and independently unreproducible. The exact next experiment is still: land and fingerprint the scheduler, pre-generate and persist identical seeds/push schedules, then evaluate zero action, brace controls, the Stage-1a checkpoint, and the candidate on calibration and held-out panels while saving one row per episode and shove.

kuds:
## Revision 3 in `80d69ce` — addendum findings accepted; the reset-validity bug is reproduced and promoted to blocking

I independently reproduced the load-bearing empirical claims before revising. Every one holds.

### Reproduced

**Reset validity — confirmed, and it is a real environment bug.**

```
seed 3077: pelvis_height 0.66230 after one zero-action step, terminated=True
scan of seeds 3042-5041: 16/2000 = 0.800% already-terminal
```

Consistent with the 43/4000 = 1.075% found over the wider range. This is the single most
consequential item in the addendum, because it is not a documentation defect — roughly 1% of
episodes are unwinnable before the policy acts, and it is the direct cause of the 39/40 panels.
Promoted to **§7.5, blocking, sequenced first**.

**Power analysis — every figure reproduces exactly.**

```
    n   cutoff   P(pass | p=0.95)   P(pass | p=0.98)
   30    30/30           21.464%            54.548%
   40    40/40           12.851%            44.570%
   80    77/80           42.845%            92.315%
  100    96/100          43.598%            94.917%

smallest n with >=80% power at p=0.95:  n=179, cutoff 168  (81.29%)
n=180 -> cutoff 169, power 80.81%
```

**Fail-open plumbing — confirmed at the two cited sites.** `jax_curriculum.py:36-37` logs a
warning and `return True` when `min_avg_reward` is absent; `StageThreshold` (`curriculum.py:79-84`)
defaults every omitted threshold permissively. Your composite-gate experiment showing
`thresholds_from_configs` discarding unknown fields and SB3 advancing on `-1e12` reward with
length `1` is exactly the failure mode that makes §2.3's removal of `min_avg_reward` dangerous
in isolation.

### Changes in revision 3

| finding | change |
|---|---|
| 1 — fail closed | New **§5.2** with your measured fail-open evidence. Versioned `gate_schema_version`/`gate_kind`, unknown kinds and fields fatal when advancement is enabled, absence allowed only in a recorded non-advancing pilot mode, config + SB3 + JAX + tests land atomically — noting the existing tests currently *assert* the permissive behaviour and must change in the same commit. |
| 2 — statistical contract | New **§2.3.1**: Clopper-Pearson named as the interval method, the cutoff/power table above, the measured fact that the current checkpoint fails half the panels, and two declared operating points. `required_consecutive` explicitly demoted to scheduler hysteresis with one predeclared held-out confirmation panel. **§3.4** now states the multiplicity rule (simultaneous bounds or select-then-confirm) and requires per-episode pair records, since marginal totals cannot reproduce a paired decision. |
| 3 — normative vs evidential | **§5** reframed around three frozen artifacts: `capability_spec` (normative, versioned), `null_manifest` (evidential), `decision_procedure` (predeclared). Baselines support the preflight and relative superiority only; they do not set capability requirements. |
| 4 — resume vs init | **§3.2** now defines `resume_same_stage` and `initialize_next_stage` as distinct modes, and narrows the promise to reproducible *stage-boundary* restart until PRNG, scheduler state, schedule position, transition count and ramp progress are persisted. |
| 5 — executable manifests | **§4** gains a versioned ordered per-species manifest with `id` / `config` / `terminal` / `legacy_alias`, task identity separated from evaluation protocol, and per-shove recovery evidence fields. |
| 6 — stale surrounding docs | **PR description rewritten** to match revision 3. **§11** now records that `TREX_REVIEW` §NS-2 still recommends the survivor-conditioned floor §5.1 rejects, and that its NS-1 correction 5 contradicts NS-2 on T-Rex — flagged for explicit supersession. |

All six specific corrections applied: bound direction (**above**, I had it inverted), collapse
fallback is `0.0` rather than undefined with the eager-arming hazard named, resolver moved
*before* stage enablement in §10, `FULL GATE` relabelled as the one-panel reward+length
aggregate it is, `perturbation_delta_v` replaced by a dimensionless multiple with derived
force/impulse persisted per species, and the impossibility claim narrowed to the undisturbed
task.

**§1.4** now carries your four-panel pooled figure — 93/160 = 58.125%, CP 50.08%–65.87%, with
per-panel reward means spanning 1884.71–2157.53 — as the empirical argument for a state-capability
gate over a reward scalar.

### On the audit

Your read is right: "all findings addressed" was too strong, and I'll drop that framing. Also,
the "CI green on `9e9924f`" line in my previous comment was written before that run finished. It
did complete green, but I asserted it without checking — in a comment about verifying claims
before making them.

### Unchanged and unresolved

The pushed-task numbers in §3.4 remain `[artifact-derived, unverified]`. The scheduler, force
conversion, registered schedule and raw per-episode outcomes are absent from the repository, and
the available checkpoint differs from the artifact behind the published `2489.65`. §11 states
this plainly rather than implying reproducibility. Your final experiment description is now the
document's §10 step 10 nearly verbatim.

No disagreement with anything in this round.

---

PR #475, opened 2026-07-30
Refactor: split the four largest shared modules into layered packages

kuds:
## Summary

Structural cleanup of `environments/shared/`. The four largest modules there held 6,268 lines between them; each is now a package split along the seams its own docstring already described. The test files that cover them are split to match, following the `test_sweep_*.py` convention the repo already had.

Public surfaces are unchanged — every package `__init__` re-exports exactly what the module exported, so no import site outside the packages moved. Two real bugs were found along the way and fixed (details below).

## Source splits

| Module | Lines | Became |
|---|---|---|
| `reporting.py` | 1,929 | `formatting`, `gates`, `csv_output`, `text_summaries`, `summaries`, `bundles`, `stage_artifacts` |
| `plant_contract.py` | 1,822 | one module per contract layer (`source_layer`, `policy_layer`, `physics_layer`, `visual_layer`) over `constants`/`errors`/`digests`/`identity`/`versions`/`introspection`, plus `manifest`, `validation`, `__main__` |
| `curriculum.py` | 1,316 | `sb3_compat`, `manager`, `schedules`, `advancement`, `early_stopping`, `checkpoints` |
| `result_bundle.py` | 1,201 | `constants`, `errors`, `naming`, `hashing`, `provenance`, `manifest`, `evidence`, `audit` |

Two notes on the layering: `_load_generated_manifest` moved down into `plant_contract/versions.py`, because `_species_entries` and `load_plant_versions` both fall back to it and leaving it in `manifest.py` made the dependency circular. And `CurriculumManager` no longer shares a module with the SB3 integration, so a JAX or notebook loop can drive the curriculum without importing SB3 at all.

`python -m environments.shared.plant_contract --check` still works via the package `__main__`.

## Test splits

| Was | Now |
|---|---|
| `test_curriculum.py` 1,728 | 6 files, largest 462 |
| `test_result_bundle.py` 1,229 | 4 files, largest 454 |
| `test_reporting.py` 1,089 | 6 files, largest 452 |
| `test_plant_contract.py` 603 | 5 files, largest 271 |

Tests moved verbatim; no assertion changed. Collected test IDs were diffed before and after for each file and are identical, with one deliberate exception: `test_parses_serialized_booleans_strictly` moved from `TestStrictRecordedGate` to a new `TestParseOptionalBool`, because it tests `parse_optional_bool` rather than the gate.

Shared fixtures were extracted rather than duplicated — `reporting_helpers.py` and `result_bundle_helpers.py` follow the existing `*_helpers.py` convention, and the two actual pytest fixtures live in a new `conftest.py` (a fixture imported into a module reads as an unused import, and `ruff --fix` would delete it).

## Bugs found and fixed

**Result-bundle write tripwire had silently weakened.** `test_identical_completed_export_is_a_write_free_noop` patches `_write_json` to prove an identical re-export performs no writes. In the single module every caller resolved it through module globals, so one patch covered all of them; after the split, `provenance.py` and `manifest.py` bound it by value and the patch stopped reaching them. Fixed by routing every internal caller through `hashing._write_json`. Confirmed by probe (all three writers intercepted before, none after) and by mutation: disabling the early return leaves the test green under the old patch target and red under the new one.

**Default provenance capture resolved the repository root one level too shallow.** `initialize_result_bundle` derived it as `Path(__file__).resolve().parents[2]` — correct in `environments/shared/result_bundle.py`, off by one once the code moved into `result_bundle/provenance.py`, where it resolved to `<repo>/environments`. `git status` reports repo-wide while `git ls-files --others` is scoped to its working directory, so an untracked file under `configs/` or the repository root set `repository_dirty` without entering `repository_patch_sha256` — materially different dirty trees could share one provenance hash. The root is now `REPOSITORY_ROOT` in `result_bundle/constants.py`, derived from a named `_SHARED_ROOT` anchor rather than a bare parent count, read through the `constants` module so there is a single authoritative binding.

Worth recording: that expression was byte-identical across the move, so an AST-level verbatim audit reported it unchanged. Only evaluating it exposes the bug. A sweep of every remaining `__file__`-relative path in the moved code found no other instance.

## Other changes

- **`harnesses/`** — `test_env_base.py`, `test_actuators_base.py` and `view_model_base.py` were hand-run smoke checks and viewers living in the importable package root under `test_` names. pytest collected their functions as tests and errored on the unfillable `env_class`/`cfg` arguments for any invocation that reached them; only `testpaths`/`norecursedirs` was hiding it. Now `harnesses/env_smoke.py`, `harnesses/actuators.py`, `harnesses/viewer.py`.
- **Python 3.12 added to the five test matrices.** `pyproject` already claimed 3.12 support and mypy already targets it, but the per-species environment tests only ran 3.11 and 3.13.
- **Intra-package I/O seams routed through their owning module** (16 of them), so each has one patch point re-read at call time. Four are deliberately left binding by name, with a comment at each: `manifest`, `provenance` and `versions` are also used as ordinary local variables in the files that import them.
- **Removed the top-level `Images/` directory** — 25 MB whose three GIFs were byte-identical to the copies under `website/static/img/` that the site and README actually reference. Nothing referred to it except the `.dockerignore` entry excluding it.
- **CI** now globs `test_plant_contract_*.py` / `test_curriculum_*.py` instead of naming the pre-split files.
- **Docs** — `TREX_LEG_FLEXING_PLAN.md` and `KNOWN_ISSUES.md` repointed at the new module locations. `docs/reviews/` and `docs/investigations/` deliberately keep their original line references: `docs/README.md` records those as point-in-time archives that are never rewritten.
- **Version** bumped `0.3.4.dev0` → `0.3.5.dev0`.

## Verification

- Test suite **1712 → 1715 passed / 102 skipped**, on both 3.11 and 3.12 locally. The three new tests are the provenance regressions.
- `ruff check`, `ruff format --check`, `mypy environments/` all clean.
- All 21 CI checks green on the head commit, including `plant-contract` (which exercises the package `__main__` against a fresh wheel install) and `test-jax-cpu` (the only job covering `save_jax_stage_artifacts`).
- Both module CLIs pass: `plant_contract --check` and `species_catalog --check`.
- All four species' `scripts/test_env.py` run end to end with output structurally identical to pre-refactor (the numeric differences are pre-existing unseeded `action_space.sample()`, confirmed by a pre-vs-pre control run).
- All 53 `environments.*` imports across the notebooks execute successfully, unchanged from before.
- Importing `environments.shared` pulls in exactly the same set of third-party packages as before — no new import cost.


environments/shared/result_bundle/provenance.py
```diff
@@ -0,0 +1,288 @@
...
+        for role, role_seed in normalized_seed_roles.items()
+        if "evaluation" in role
+    }
+    normalized_plant_identity = _normalize_plant_identity(plant_identity, species=species)
+
+    run_path = Path(run_dir)
+    run_path.mkdir(parents=True, exist_ok=True)
+    provenance_path = run_path / DEFAULT_PROVENANCE_NAME
+    if repository_root is None:
+        repository = Path(__file__).resolve().parents[2]
```
kuds: [P2] Restore the repository-root depth for default provenance capture

This expression was correct while it lived in `environments/shared/result_bundle.py`, but after moving to `result_bundle/provenance.py`, `parents[2]` resolves to `<repo>/environments` rather than `<repo>`. `_repository_state()` runs `git status` repo-wide, while `git ls-files --others` is scoped to its working directory, so untracked files under `configs/` or the repository root can set `repository_dirty` without being included in `repository_patch_sha256`. Distinct dirty trees can therefore receive the same provenance hash. Please use `parents[3]` (or derive the top level with `git rev-parse --show-toplevel`) and add a regression test with a root-level untracked file.

environments/velociraptor/scripts/test_actuators.py
```diff
@@ -4,15 +4,15 @@
 Useful for verifying joint ranges and actuator gains.
 
 Usage:
-    python test_actuators.py
+    python run_actuator_test.py
```
kuds: [P3] Keep the usage command aligned with the executable filename

Only the shared function was renamed; this executable is still `test_actuators.py`, so `python run_actuator_test.py` fails because that file does not exist. The same stale command appears in the T-Rex and Dibothrosuchus scripts and in `shared/harnesses/actuators.py`. Please restore `python test_actuators.py` or rename and update every caller/reference.

CHANGELOG.md
```diff
@@ -87,6 +87,46 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   effective hunting curriculum. 12M matches the brachiosaurus stage-3
   budget.
 
+- **The five largest `environments/shared` modules are now packages**, each
```
kuds: [P3] Correct the package count

This section enumerates four converted modules—`reporting`, `plant_contract`, `curriculum`, and `result_bundle`—rather than five.

kuds (commented):
Re-reviewed the updated head (`906520a0`). The new result-bundle write-tripwire commit looks sound: the targeted audit/provenance/manifest/reporting suite passes, and the updated Python CI run is green.

One contributor command outside the changed-file diff is still stale: `docs/PLANT_CONTRACT.md:67` invokes the deleted `environments/shared/tests/test_plant_contract.py`. Please change it to `environments/shared/tests/test_plant_contract_*.py`, matching CI.

The remaining findings are attached inline.

kuds:
All four items addressed in `6cca926`.

**P2 — provenance repository root.** Confirmed and fixed. I verified the mechanism you described rather than assuming it: `git status` reports repo-wide while `git ls-files --others` is scoped to its cwd, so a root-level untracked file flipped `repository_dirty` without entering `repository_patch_sha256`.

One deliberate departure from your suggestion, worth flagging: rather than `parents[3]` or `git rev-parse --show-toplevel`, the root is now `REPOSITORY_ROOT` in `result_bundle/constants.py`, derived from a named `_SHARED_ROOT` anchor. Reasoning — `parents[3]` is correct today but re-arms the identical trap on the next move, and shelling out to git would make the default path depend on git being present and the tree being a checkout, which it currently does not. The named anchor matches what `plant_contract/constants.py` already does, so both packages now derive paths the same way.

Two regression tests, both mutation-checked (restoring the old depth turns the first red):
- `test_repository_root_resolves_to_the_repository` — pins `REPOSITORY_ROOT` to the directory holding `pyproject.toml`
- `test_untracked_file_at_repository_root_enters_the_patch_hash` — builds a real git repo, asserts a root-level untracked file changes the patch hash, and asserts a too-deep root *cannot* see it

I also swept every remaining `__file__`-relative path in the moved code. Two exist, both now anchored, all derived paths evaluate correctly — no other instance.

Worth recording why this got past the earlier verbatim audit: the expression was byte-identical across the move, so an AST-level diff reports it unchanged. Only evaluating it exposes the bug.

**P3 — three stale references.**
- `docs/PLANT_CONTRACT.md:67` → `test_plant_contract_*.py`, matching CI.
- CHANGELOG "five largest" → "four".
- Actuator usage lines restored to `python test_actuators.py` in all three species scripts and `harnesses/actuators.py`. This was a rename script over-reaching into docstrings that name the file rather than the function. I added a check comparing every species script's usage line against its own filename; the rest were already correct.

Suite 1712 → 1714 passed / 102 skipped on 3.11 and 3.12; ruff, ruff-format and mypy clean.

---

kuds (commented):
The production fix is correct and the latest CI run is green. I found one non-blocking gap in the regression coverage, attached inline.

environments/shared/tests/test_result_bundle_provenance.py
```diff
@@ -0,0 +1,272 @@
...
+    module moved into the package — it resolved to `<repo>/environments`. The
+    expression stayed byte-identical through the move, so an AST diff could not
+    see it; only evaluating it can.
+    """
+    assert (result_bundle.constants.REPOSITORY_ROOT / "pyproject.toml").is_file()
+    assert (result_bundle.constants.REPOSITORY_ROOT / "environments").is_dir()
+    assert result_bundle.constants.REPOSITORY_ROOT.name != "environments"
+
+
+def test_untracked_file_at_repository_root_enters_the_patch_hash(tmp_path: Path) -> None:
```
kuds, line 277: [P3] Pin the initializer's default-root wiring

This test passes both roots directly to `_repository_state`, while the preceding test only validates `constants.REPOSITORY_ROOT`; neither calls `initialize_result_bundle` with `repository_root` omitted. Consequently, restoring the original wrong assignment in `provenance.py` leaves both tests green—I reproduced that by forcing `provenance.REPOSITORY_ROOT` to `<repo>/environments`. Please exercise the public initializer's default path, for example by spying on `_repository_state` and asserting it receives `constants.REPOSITORY_ROOT`, so the original wiring regression cannot recur unnoticed.

kuds:
Good catch — reproduced it before fixing. Restoring `Path(__file__).resolve().parents[2]` inside the initializer left all 14 tests green, exactly as you described.

The root cause is the same by-value trap this branch has been chasing elsewhere: `provenance.py` held its own `REPOSITORY_ROOT` binding imported from `constants`, so pinning `constants.REPOSITORY_ROOT` could not observe a change to how `provenance.py` used it. Fixed in `a323c28` on both sides:

- `provenance.py` now reads `constants.REPOSITORY_ROOT` through the module, so there is one authoritative binding rather than two.
- `test_initializer_defaults_to_the_repository_root` spies on `_repository_state` and asserts it receives `constants.REPOSITORY_ROOT`, exercising the public initializer with `repository_root` omitted.

The two failure modes are now caught by distinct tests, each mutation-checked:

| mutation | test that fails |
|---|---|
| wrong wiring in the initializer | `test_initializer_defaults_to_the_repository_root` |
| wrong depth in the constant | `test_repository_root_resolves_to_the_repository` |

Suite 1714 → 1715 passed / 102 skipped on 3.11 and 3.12; ruff, ruff-format and mypy clean.

Unrelated to the review, the same push bumps the package version `0.3.4.dev0` → `0.3.5.dev0` (and the CHANGELOG heading that labels it) at the author's request.

---

PR #484, opened 2026-08-03
Make stage 1 certifiable: enforce the declared gate, stop the backstop arming on the untrained policy

kuds:
Two independent reasons T-Rex stage 1 has never produced a certified checkpoint, plus the lessons written up. Both defects were found by reading the last two runs' artifacts against each other; both are reproduced in tests before being fixed.

---

## 1. The notebook never enforced the declared gate

`notebooks/sb3_training.ipynb` computed `publication_gate_passed` from its own inline checklist over `min_avg_reward` / `min_avg_episode_length` / `min_avg_forward_vel` / `min_success_rate`. It never called `CurriculumManager.should_advance`, which is where `stance_quality/v1` is evaluated — `_build_core_callbacks` constructs no `CurriculumCallback`, so **on the Colab path the stance gate was dead code**.

Retiring `min_avg_episode_length` from the trex 1a config then made the checklist's only other criterion vacuous (`.get(..., 0.0)`), leaving *reward alone* — the one threshold the zero-action statue clears by 68%.

| run | stage 1 recorded | its own stance report said |
|---|---|---|
| `20260802_203215` | `stage_passed = True` | `GATE: FAIL` — `mean_unsupported_duty 0.2120 > 0.0200` (10.6×) |
| `20260803_012355` | `stage_passed = True` | `GATE: FAIL` — `full_horizon_fraction 0.6500 < 0.9500` |

**Fix.** The rule now exists once, as `reporting.gates.evaluate_stage_gate`, dispatched on the declared `gate_kind` and called from `generate_stage_artifacts` — the entry point the notebook and the sweep trial worker already share. The notebook's checklist is **deleted**, not reimplemented. Fail-closed at every branch: undeclared kind, unknown kind, `none/v1`, missing stance panel, a verdict scored for a different gate, and a non-boolean verdict all return `False` with a stated reason.

```
OLD notebook checklist -> passed = True   criteria checked: 1
NEW shared gate        -> (False, ['mean_unsupported_duty 0.2120 > 0.0200',
                                   'unsupported_duty_ucb 0.2153 > 0.0200'])
NEW, report missing    -> False
```

Stages 2 and 3 are unchanged: thresholds still apply only when declared, velocity and success floors only when positive.

**Fail-open holes found probing the new code**, all closed with regression tests asserting the observed wrong answer inverted:

| probe | observed | now |
|---|---|---|
| NaN metric vs `min_avg_reward = 100` | `(True, [])` — `nan < threshold` is `False` | fails, "no reward measurement available" |
| `reward_and_length/v1` declared, no threshold set | `(True, [])` — vacuously-true empty conjunction | fails, "checks nothing" |
| `failures` given as a bare string | shredded into 13 one-character reasons | kept intact |

---

## 2. The collapse backstop armed on the untrained policy

With `home-keyframe-residual/v1` and `log_std_init = 0`, action = 0 commands the nominal stance, so an **untrained** T-Rex already scores near the reward optimum. Run `20260803_012355` peaked at **2469.4 on its second evaluation** (100k steps), clearing the `0.45 × 3271.8 = 1472.3` arming floor on initialisation. The exploration dip then read as a collapse: training stopped at **1.45M of a 10M budget**, and stages 2 and 3 spent 9¼ hours on the near-statue checkpoint it left.

Decomposing that checkpoint as a two-point mixture of survive/fall episodes (from its stance panel: mean 2319.7, sd 1263.7, 26 of 40 full-horizon) gives **survive = 3247.0**, within **0.76%** of the zero-action statue's 3271.8. Cross-checked against the 100k evaluation: implied survival 70.65%, predicted sd 1206.4 vs 1195.52 observed — 0.91% apart. It is the statue, on whichever episodes it doesn't fall over.

**No `peak_floor` value fixes this.** On stage 1 "already good" and "hasn't started learning" are the same number — which is also why the two absolute floors documented in `collapse_settings_from_config` never armed at all.

**Fix.** `collapse_peak_warmup_timesteps` — rolling windows containing any evaluation before it cannot **set** the peak. Eligibility is by window *start*, so no surviving window straddles the boundary; a median absorbs one contaminating sample out of five, not five. Disarmed while nothing qualifies, and disarmed if evaluation timesteps are unreadable rather than aborting a multi-hour run on a signal it cannot read. Expressed in timesteps rather than reward, so unlike an absolute floor it survives a reward-function edit.

Deliberately **not** a tightening of `drop_fraction`/`patience`, which `collapse_settings_from_config` documents at length as aborting healthy runs.

Defaults to `0.0` — verified across every config in the tree, only trex 1a sets it, to 2.5M. Replaying the run's shape through the callback stops it at **exactly 1,450,000** without the warm-up and never arms with it.

⚠️ The 2.5M value is `[inferred]` from three runs' breakthrough timings (2.2–2.3M, 2.8M, 3.3M, 4.55M), **not** from a simulated sweep over the run's own 29-evaluation series. The config and §12.4 both say so.

---

## 3. Lessons learned — `docs/STAGE1_SPLIT_PLAN.md` §12

- **12.1** A gate the trainer never calls is not a gate. Three copies of the advancement rule existed; the one that ran was the one nobody had updated. Ask what would have to be *deleted* for a gate to stop being consulted.
- **12.2** On stage 1 the untrained policy is already near the reward optimum (survive 3247.0 vs statue 3271.8), so no reward-referenced threshold separates competence from initialisation. This one fact explains three separate historical failures.
- **12.3** A robust statistic defends against an outlier, not a regime. The rolling median of 5 absorbs one inflated evaluation; the statue regime sits **5.2σ** above the floor, so five consecutive draws from it have a median far above the floor too.
- **12.4** The 6M run was lucky, not protected — see the answer below.
- **12.5** Reproduce the failure before claiming the fix, or a test cannot distinguish "I fixed it" from "I disabled it".
- **12.6** Fail-open defects cluster around *absence* — NaN, empty conjunction, missing panel. Enumerate every path that returns "pass" without having compared anything.
- **12.7** What this does **not** explain: the two runs failed the gate on opposite criteria (duty 0.2120 at full-horizon 1.0000; full-horizon 0.6500 at duty 0.0000), and neither result shows whether a policy satisfying both exists under the current reward.

### Why the 6M run didn't hit early stopping and the 10M run did

The configs differ by **exactly one line** — `timesteps = 6000000` → `10000000` (`git diff 2d032ce..8f0f7a9`). Only the LR schedule reads it: `linear_schedule` is parameterised on `progress_remaining`, so the 10M run holds a 0.45% higher LR at 100k and 7.7% higher by 1.45M. `ent_coef_decay_timesteps` is absolute; nothing else in the stage is a function of the budget.

So the 6M run was not protected by anything the 10M run lacked. Since the statue regime sits 5.2σ above the arming floor and both runs share a seed and an initialisation, both almost certainly armed. What differed was whether the dip produced `patience = 10` **consecutive** sub-threshold evaluations — it did in the 10M run, and did not in the 6M one, whose sibling `20260801_021545` put 45 of 94 pre-peak evaluations below half their running peak, frequent but interspersed with recoveries that reset the counter.

That is the argument for the warm-up independent of which run drew badly: a backstop whose arming depends on a noise-dominated early evaluation converts run-to-run variance into a categorical outcome.

---

## 4. Species catalog regeneration

`configs/trex/stage1_balance.toml` went 6M → 7M → 10M across three commits without regenerating the catalog those values are baked into, so `check_catalog` reports it stale and both catalog tests fail. Reproduced on a clean `main` worktree with none of this branch's code present — it fails CI on any branch cut from `main`.

## Testing

- `environments/shared/tests`: **1,865 tests, exit 0** on the gate fix; re-running with the warm-up.
- 26 new tests across `test_reporting_gates.py`, `test_reporting_stage_artifacts.py` and `test_curriculum_early_stopping.py`.
- `ruff check` / `ruff format --check` clean across `environments/`; `mypy` clean on the changed modules.

## Known remaining gap

`sweep/ray_tune.py:1097` keeps a third private copy of the reward/length/vel/success checklist via `_evaluate_curriculum_gate`. It ranks finished trials rather than gating advancement, so it needs different handling and is deliberately out of scope.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017svujdxTHg8fWZunY638hU


kuds:
Closing as superseded by #485.

#485 contains all five commits from this branch unchanged — cherry-picked, and the resulting tree is byte-identical to `claude/review-prs-plant-validation-8lm5od`. Nothing here is dropped. It adds seven commits fixing gaps found while reviewing this PR, and then while adversarially reviewing that review.

The substantive reasons to move rather than merge this one:

**The warm-up value in this description is stale.** §2 and the ⚠️ note say `2.5M` and `[inferred]`. The final commit here (`89c0877`) re-derived it to **1.0M**, measured against the run's own evaluation series, and the config, CHANGELOG and §12.4 all say 1.0M. Merging with this body would record the wrong number in the merge commit.

**§12.4's explanation of the 6M run is contradicted by that run's data.** This description argues both runs "almost certainly armed" and the 6M one survived because its dip never produced ten consecutive sub-threshold evaluations. Reading `20260802_203215`'s `evaluations.npz` (120 evaluations) says otherwise: its first rolling-median window is **812.4**, 45% *below* the 1472.3 floor rather than 35 points below it, and it records **zero** consecutive sub-threshold evaluations across the whole run. It never armed early, and `patience` was never the deciding factor. What differed was the initialisation spike — first evaluation 812.4 against 2007.3, 2.47× at the same seed. #485 corrects §12.3 and §12.4 accordingly. The correction strengthens the case for the warm-up: arming turns on how large an untrained policy's first evaluations happen to be, which no threshold can control.

**Three fail-open holes remained.** `jax_eval.check_stage_gate` had the identical NaN hole this PR closed in `reporting.gates` — `(True, [])` for an evaluation that measured nothing — on the path that gates JAX stages 2 and 3, with `jax_setup` writing the verdict straight into `publication_gate_passed`. The other two were in #485's own additions, found by probing them the way §12.6 prescribes.

**The notebook enforcement this PR depends on was pinned by nothing.** Deleting the three enforcement blocks from `sb3_training.ipynb` leaves the full `environments/shared/tests` suite at exit 0 — including all four tests that already parse that notebook. Every run would advance on a failed gate with CI green. That is §12.1's own lesson reappearing one level up, in the artifact the lesson was written about.

**A certified stage 1 still could not be published.** `result_bundle.evidence` refused any complete bundle declaring `stance_quality/v1`, since `evaluation_selected.csv` carries no duty column. Correct, and it also made the milestone unreachable — the refusal fires only after all three stages pass, so the first genuinely successful run would train for ten hours and fail at finalisation. Visible on run `20260803_012355`, whose run-level `collected_results.csv` and `artifact_manifest.json` are frozen at stage 2 while stage 3's own artifacts were written in full. #485 persists the panel's per-episode duties and has the auditor re-derive the verdict from them.

The analysis in this PR held up everywhere it was checkable against the Drive artifacts — the two runs' stance reports, the `publication_gate_passed = True` recorded beside `GATE: FAIL`, `best_eval_step = 100000`, and the two-point mixture decomposition (survive 3247.0, implied survival 70.65%, predicted sd 1206.4 vs 1195.52 recorded) all reproduce exactly.

---

Issue #486, opened 2026-08-04
T-Rex stage 1: policy converges to a locked 16.7 Hz bounce — unsupported duty pinned at exactly 1/6, 8.3x the gate ceiling


kuds:
Run `20260804_143747` is the first T-Rex stage 1 to train its full 10M budget and reach a real gate verdict. It **fails on unsupported duty alone**, and the per-episode evidence shows why in a way the previous runs never made visible: the policy has converged to a **deterministic limit cycle with both feet airborne for exactly one control step in every six**.

Run at merge commit `27d366a` (#485). Two of three criteria pass comfortably; the third is not close and is not moving.

---

## 1. The verdict

`stage1/stance_gate_report.txt`, 40 episodes at seeds 3042–3081, `settle_steps 200` (duty over the remaining 800 steps):

| criterion | measured | threshold | |
|---|---|---|---|
| `full_horizon_fraction` | **1.0000** | ≥ 0.95 | pass |
| reward (rail) | **2686.9 ± 28.3** | ≥ 1950 | pass |
| `mean_unsupported_duty` | **0.1668** | ≤ 0.02 | **fail — 8.3×** |
| `unsupported_duty_ucb` | **0.1670** | ≤ 0.02 | **fail — 8.4×** |

Supporting: `bilateral_support 0.8332` (statue 0.998), `single_support 0.0000`, terminations `{'truncated': 40}` — **zero falls in 40 episodes**.

## 2. The key finding: it is not jitter, it is a locked cycle

Across all 40 independent episodes, `unsupported_duty` takes only **two distinct values**:

| duty | airborne steps (of 800) | episodes |
|---|---|---|
| 0.16625 | **133** | 22 |
| 0.16750 | **134** | 18 |

Between-episode standard deviation is **0.00063** — a spread of ±1 step in 800, across 40 different seeds. That is not stochastic foot-shuffling; it is a phase-locked oscillation.

```
measured duty   0.166813
1/6             0.166667      ratio 1.00088
```

Episode = 1000 steps = 10.00 s sim time → 0.01 s per control step (100 Hz). One airborne step per 6-step cycle → a **~16.7 Hz vertical oscillation** in which both feet break contact together at the top of each cycle.

`single_support_duty` is **exactly 0.0000**. The feet never leave or land independently. This is a bounce, not a gait and not a weight shift.

[...]

PR #488, opened 2026-08-05
Decay trex stage-1 entropy to zero: the policy was being told to shake its legs by 20° for the whole run

kuds:
Two-line config change for issue #486, with a measured mechanism and a falsifiable prediction attached.

```diff
-ent_coef_end = 0.001
-ent_coef_decay_timesteps = 3000000
+ent_coef_end = 0.0
+ent_coef_decay_timesteps = 7000000
```

## The mechanism

Run `20260804_143747` trained the full 10,002,432 steps and finished at **2686.9** against the zero-action statue's **3271.8** — 588 *below* the do-nothing policy — with unsupported duty pinned at 0.1668, 8.3× the ceiling.

`ent_coef` reached its 0.001 floor at 3M and held there for the remaining 7M. Across 21 action dims that residual bonus holds the policy std at equilibrium ~0.375:

| step | `algo_std` |
|---|---|
| 2.1M | 0.879 |
| 4.6M | 0.554 |
| 5.9M | 0.469 |
| 8.4M | **0.375** |

Still falling ~0.01 per 1M at the end — it was never going to converge inside the budget.

**What that std means physically.** Converted through `_scale_action`'s residual mapping and this model's actual ctrlranges:

| joint | 1σ command noise |
|---|---|
| hip pitch | **±24.4°** |
| knee | **±18.8°** |
| ankle | **±18.8°** |
| mean, major leg joints | **±20.6°** |

Resampled every control step, at 100 Hz, for the entire run.

**A policy cannot learn to hold a pose it is being commanded to shake by 20°.** PPO optimises expected return *under* its own action noise, so the mean action it converges to is not the deterministic optimum — it is whatever survives the tremor. In that run: a 16.7 Hz limit cycle, both feet leaving and landing together, `single_support_duty` exactly **0.0000**.

## Why this is a convergence failure, not a reward problem

Measured locally, not quoted from a run:

```
$ python environments/shared/scripts/stance_gate_report.py trex --stage 1 --zero-action --episodes 8

reward                    3274.4 +/- 7.6
mean_unsupported_duty     0.0000   (<= 0.0200)
  bilateral support       1.0000
```

Under `home-keyframe-residual/v1`, **`action = 0` *is* the home keyframe**. The gate's 0.02 ceiling is achievable under the *current* reward, by the most trivially reachable policy in the space, and the reward prefers it by 588 points on every major term.

That also makes the usual objection to zeroing entropy — premature convergence to a bad optimum — unusually weak here. The good optimum is trivial to reach, and the observed failure was never converging at all.

## The choices

**0.0, not a smaller floor.** Any floor sets a non-zero std equilibrium. Reaching std ~0.05 (2.75°/joint, a plausible standing policy) needs the bonus to go away, not to get smaller.

**7M, not 3M.** The 3M anchor was chosen for a *6M* stage (~half of it) and was never revisited when stage 1 became 10M — so decay finished in the first third and the floor governed the rest. Keeping decay live through 7M means entropy is still falling while the policy consolidates.

**Early exploration is deliberately unchanged.** `ent_coef` still starts at 0.005 and the first 3M now decays *more slowly* than before, so nothing about the early regime this stage depends on is tightened:

```
        0 steps: 0.005000
1,000,000 steps: 0.004286
3,000,000 steps: 0.002857   <- old config was pinned at 0.001 from here on
5,000,000 steps: 0.001429
7,000,000 steps: 0.000000
```

## Falsifiable prediction — check at ~4M, not at the end

If the hypothesis is right, by roughly 4M steps:

- **`algo_std` well under 0.2** — it was 0.554 at 4.6M last run
- **`unsupported_duty` below 0.28** — it was 0.282 at 4.34M last run

If either tracks the old curve, **this hypothesis is wrong** and the next suspect is the learning rate (3e-5 → 1e-5 is very low for escaping a basin). Kill the run at that point rather than spending the remaining five hours.

Both are visible live: `algo_std` and `unsupported_duty` are already logged into `diagnostics.npz` every rollout.

## Interaction with #487

Independent — this branch is cut from `main` and touches only `configs/trex/stage1_balance.toml` and the CHANGELOG. But they compose well: with #487 merged, a run that is still below the statue at 3.5M says so in the log, which is the same "kill it early" signal from a different angle. Either order works.

## Testing

Config-only, no code paths changed.

- `_maybe_ent_coef_decay_callback` gates on `is None`, not truthiness, so `0.0` still builds the callback — checked, since a falsy-zero bug there would have silently disabled decay entirely
- Full `environments/shared/tests` plus `environments/trex/tests`: **exit 0**
- `ruff check` clean; `check_catalog()` reports up to date (no regeneration needed)
- Resolved schedule verified end to end from `load_stage_config`

## What this does not claim

That it will pass the gate. It is one hypothesis with one measured mechanism, and the prediction above is there so a failure is cheap and legible rather than another 8-hour ambiguity. If std collapses and duty *still* plateaus above 0.02, that is a genuinely informative negative — it would rule out exploration noise and point at the objective PPO actually optimises (discounted return under `gamma 0.98`) rather than the episode return.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01MnEFdfBDkHAnVnxExdrzaG)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] — Reproducible Runs & Velociraptor Stage-1 Diagnosis (v0.3.6)
 
+### Changed
+- **T-Rex stage 1 entropy now decays to zero, over 70% of the budget rather
+  than 30%** (`ent_coef_end` 0.001 → 0.0, `ent_coef_decay_timesteps` 3M → 7M).
+  Run `20260804_143747` trained the full 10M and finished at 2686.9 against
+  the zero-action statue's 3271.8 — **588 below the do-nothing policy** — with
+  unsupported duty pinned at 0.1668, 8.3× the gate ceiling (issue #486).
+  The cause is measured, not guessed: `ent_coef` reached its 0.001 floor at 3M
+  and held for the remaining 7M, which across 21 action dims holds the policy
+  std at equilibrium ~0.375 (`algo_std` 0.47 at 5.9M, 0.375 at 8.4M). Through
+  `_scale_action`'s residual mapping and this model's ctrlranges that std is a
+  1σ command noise of **24.4° at the hips and 18.8° at knee and ankle**, mean
+  **20.6°** across the major leg joints, resampled every control step at
+  100 Hz. A policy cannot learn to hold a pose it is being commanded to shake
+  by 20°; PPO optimises expected return *under* that noise, so it converged to
+  a 16.7 Hz limit cycle that survives its own tremor rather than to a still
+  stance.
+  The target is known reachable and known still: `action = 0` **is** the home
+  keyframe under `home-keyframe-residual/v1` and scores 3274.4 ± 7.6 with duty
+  0.0000 (measured locally with `stance_gate_report.py trex --stage 1
+  --zero-action`). So this is a convergence failure, not a reward-shaping one,
+  and the usual risk of zeroing entropy — premature convergence to a bad
+  optimum — is unusually low here.
+  Early exploration is deliberately unchanged: `ent_coef` still starts at
+  0.005, and the first 3M now decays more slowly than before. The 3M anchor
+  was sized for a 6M stage and never revisited when stage 1 became 10M.
+
 ### Fixed
 - **A stance-gated run that passed every gate still could not publish.**
   `result_bundle.evidence` refused any complete bundle whose stage declares
--- a/configs/trex/stage1_balance.toml
+++ b/configs/trex/stage1_balance.toml
@@ -144,18 +144,48 @@ gamma = 0.98                            # Setting 4 sweep: 0.9797. Previous 0.99
 gae_lambda = 0.95
 clip_range = 0.2
 ent_coef = 0.005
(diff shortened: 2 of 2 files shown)


kuds:
Closing — consolidated into #487 at the author's request.

Commit `c980b86` was cherry-picked onto #487's branch as `5336563`, so the entropy change is unchanged in content: `ent_coef_end` 0.001 → 0.0 and `ent_coef_decay_timesteps` 3M → 7M, with the same rationale and the same falsifiable 4M prediction (`algo_std` well under 0.2, `unsupported_duty` below 0.28).

The only edit was resolving a CHANGELOG collision where both branches added a section under the same `[Unreleased]` anchor — both sections are kept, `### Added` for the baseline watch and `### Changed` for the entropy decay.

#487 now carries both, which is the better pairing anyway: the config change is the hypothesis for why run `20260804_143747` failed, and the baseline watch is what makes the next run cheap to judge whether or not that hypothesis holds.

Re-verified on the consolidated branch: full `environments/shared/tests` + `environments/trex/tests` exit 0, `ruff` and `mypy` clean, and `load_stage_config('trex', 1)` returns `ent_coef_end = 0.0`, `ent_coef_decay_timesteps = 7000000` and `baseline_warn_after_budget_fraction = 0.35` together.

---

